PR #2, opened 2019-01-20
Open files with default system application by default

### LINKED ISSUE #1: Feature request: Open files with xdg-open by default
When trying to open a file that doesn't have an entry in the `mimetype.toml`, joshuto outputs `Don't know how to open file :(`, but I think it would be good to instead open the file with `xdg-open` on Linux and `open` on Mac.

Also, just wanted to say this project looks really exciting and great so far!

cjbassi:
Closes #1


kamiyaa:
Thanks! Merged

PR #7, opened 2019-01-31
Refactor config parsing and add ./config to config hierarchy

cjbassi:
So currently joshuto doesn't work unless the config files are copied from `./config` to `~/.config/joshuto`, since no keybinds are enabled. This PR loads default config files from `./config` if there isn't one in `~/.config/joshuto/`. I also refactored out the `read_config` function from all the config parsing files in the process.

edit: This also adds the `search_config_hiearchy` function, which can be expanded in the future if other config file locations are desired.

### DIFF
--- a/src/joshuto/config.rs
+++ b/src/joshuto/config.rs
@@ -8,3 +8,36 @@ pub use self::keymap::JoshutoKeymap;
 pub use self::mimetype::JoshutoMimetype;
 pub use self::theme::JoshutoTheme;
 pub use self::theme::JoshutoColorTheme;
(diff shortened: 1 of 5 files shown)


kamiyaa:
Thanks! Although, I don't think `./config` will be used much? Probably more for testing.
Good work on the code reduction though!

cjbassi:
Yah mainly just for development. I wonder if there's a better way to implement it then, like somehow checking if we're running the program through cargo, since we don't want to be randomly checking for `./config` folders.

PR #9, opened 2019-01-31
[WIP] Migrate to Rust 2018

cjbassi:
This will likely have merge conflicts with some of the other PRs, so I think it would be best to merge those first and then we can come back and fix this


kamiyaa:
I am holding back on merging this currently because of https://github.com/jeaye/ncurses-rs/issues/180 causing unicode to not display properly.

cjbassi:
Doesn't rust2018 only require 1.31.0?

kamiyaa:
Is it okay that I close this? There are conflicts so I added Rust 2018 to dev branch because there seems to be a fix for `ncurses-rs` on rust-1.32+

cjbassi:
Yah, glad 1.32 is working now. And I realized you can just do `cargo fix --edition` which applies most of these changes anyway.

PR #10, opened 2019-01-31
[WIP] Replace delete_files command with trash_files

cjbassi:
So let me know if you have any thoughts on this, but this PR basically renames the `delete_files` command to `trash_files` and changes its behavior so it moves the files to the XDG defined trash folder instead of `rm`ing them. I think in the age of the Linux desktop, `rm`ing files is pretty outdated and it's much better to use the trash, at least for most tasks. If you think people would want the `delete_files` command too, then we could add that back in this PR. But either way, I think `trash_files` would be a better default to use in the default config instead of the `delete_files` command.

### DIFF
--- a/config/keymap.toml
+++ b/config/keymap.toml
@@ -118,11 +118,11 @@ args = [ "overwrite" ]
 
 [[mapcommand]]
 keys = [ "d", "D" ]
-command = "delete_files"
+command = "trash_files"
 
 [[mapcommand]]
 keys = [ "Delete" ]
-command = "delete_files"
+command = "trash_files"
 
 
 [[mapcommand]]
--- a/src/joshuto/command.rs
+++ b/src/joshuto/command.rs
@@ -35,8 +35,8 @@ pub use self::file_operation::CutFiles;
 pub use self::file_operation::CopyFiles;
 pub use self::file_operation::PasteFiles;
 
-mod delete_file;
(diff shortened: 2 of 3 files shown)


LovecraftianHorror:
Just weighing in that I prefer just rm ing files so that I never have to clear out my trash. I guess if files get trashed then i can just change the trash folder to /tmp or something along those lines.

kamiyaa:
In either case, we can use `open_with` and pass in `rm`, `gio trash`, etc.
But I do think trashing files is probably safer, even though I personally just `rm`.

kamiyaa:
I am refraining myself from using `fs_extra::move*` currently because it causes unnecessary copying. Across filesystems, it works well, but on the same filesystem that only requires a rename becomes a copy/delete operation that take quite a long time when operating on large files.

cjbassi:
Yah that makes sense actually. With ranger, I just do `map dD shell trash-put %s` which has worked well for me. Maybe just throw that in the wiki or something. Plus, I realized there's a fair bit of logic required to trash files, so calling trash-cli is actually a more robust method.

cjbassi:
Thanks for the feedback on this. Think its fair to close this now.

Also, @LovecraftianHorror, just a side note but I just found [autotrash](https://github.com/bneijt/autotrash) which autocleans out your trash for you. Using `/tmp` is also a good idea too like you said.

PR #13, opened 2019-01-31
Fix extra slash in current path when in root folder

cjbassi:


kamiyaa:
Thanks! Probably should've merged them in order so I get less conflicts.
This was a known issue, but I got too lazy :p

PR #14, opened 2019-01-31
Add tilde_in_titlebar option

cjbassi:


cjbassi:
Fixed merge conflicts.

cjbassi:
Also, should we use the same setting names in joshuto as in ranger, or not really? Cause in ranger, this option is called `tilde_in_titlebar`.

kamiyaa:
Nah it's okay. Even though this is like a ranger clone, I don't really intend on keeping backwards compatibility or anything. I am just busy with school currently.

kamiyaa:
Thanks!

PR #18, opened 2019-01-31
Run cargo fmt and reformat imports

cjbassi:
So I noticed you have your own style going, but I was wondering if we could switch over to using `cargo fmt` since it formats things really nicely and conveniently and it's more welcoming to new contributors. I currently use format-on-save with vscode so I don't even have to worry about it. Most editors should be able to do that I think.

I also cleaned up the import formatting cause I think it looks nice :D.

This PR also conflicts with all the others, so we should wait till those are merged to fix up and merge this too.


kamiyaa:
Yes, I am fine with switching to a guideline everyone follows.
Can you fix the conflicts? I'll merge it after its conflict free.

cjbassi:
Should be good now.

kamiyaa:
Thanks!

PR #21, opened 2019-01-31
Cleanup config hierarchy; Only add ./config in debug mode

### LINKED ISSUE #19: Only read from ./config if in debug mode
https://stackoverflow.com/questions/39204908/how-to-check-release-debug-builds-using-cfg-in-rust

cjbassi:
Close #19


kamiyaa:
Thanks, good work!

PR #23, opened 2019-02-15
Refactoring

cjbassi:
So I decided to refactor some of the config stuff I added and I ended up refactoring everything... :D. Let me know if you have any concerns about any changes. In particular, I changed a fair bit of the project layout to match some of the popular rust projects like ripgrep and alacritty and it feels a lot more idiomatic imo.

And I'm planning on working on more features now that I've gotten this out of the way. I'll probably open a few issues for features that should be added and work on those.


kamiyaa:
Wow, looks good!
Thank you for the contributions!

PR #27, opened 2020-06-29
Fix bitwise-and build issues on macOS

### LINKED ISSUE #22: Latest Build Fails to Compile
When I run

`cargo build`

I get [this](https://gist.github.com/gottaStartNow/1585579218e48a3e9e05a9779c8705e1) I am on the latest version of Rust from RustUp and the latest MacOSX. This command is being run in iTerm2.

If I had to take a wild guess as to why this is happening I would say it is due to not using the latest Rust as is mentioned in [this](
https://github.com/kamiyaa/joshuto/pull/9) pull request.

austinhyde:
I was seeing build errors due to attempting to bitwise-and a u32 and a u16:

```
error[E0308]: mismatched types
 --> src/util/unix.rs:6:50
  |
6 |     LIBC_PERMISSION_VALS.iter().any(|val| mode & *val != 0)
  |                                                  ^^^^ expected `u32`, found `u16`

error[E0277]: no implementation for `u32 & u16`
 --> src/util/unix.rs:6:48
  |
6 |     LIBC_PERMISSION_VALS.iter().any(|val| mode & *val != 0)
  |                                                ^ no implementation for `u32 & u16`
  |
  = help: the trait `std::ops::BitAnd<u16>` is not implemented for `u32`
```

I fixed this by simply up-converting `*val` to a `u32`.

The widening won't have any effect on functionality because we don't care about any of the higher bits.

(probably) Fixes #22

### DIFF
--- a/src/util/unix.rs
+++ b/src/util/unix.rs
@@ -3,7 +3,7 @@ use std::path::Path;
 pub fn is_executable(mode: u32) -> bool {
     const LIBC_PERMISSION_VALS: [libc::mode_t; 3] = [libc::S_IXUSR, libc::S_IXGRP, libc::S_IXOTH];
 
-    LIBC_PERMISSION_VALS.iter().any(|val| mode & *val != 0)
+    LIBC_PERMISSION_VALS.iter().any(|val| mode & (*val as u32) != 0)
 }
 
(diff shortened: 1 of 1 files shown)


kamiyaa:
Looks good to me! Thanks!

Issue #3, opened 2019-01-22
Implementation of Devicons?


gottaStartNow:
I am a long time user of Ranger and I love it but I always have to install devicons to make it more readable, https://github.com/alexanderjeurissen/ranger_devicons, can this be built into joshuto?

kamiyaa:
I will probably get around to this. Currently busy with school.
Feature-wise, right now I'm more focused on getting things to work with wide characters and readline like input support.

gottaStartNow:
Thank you for the update. I look forward to it.

gottaStartNow:
So it won't automatically populate the icon type next to the end of the file name? There are dozens of characters and probably many more. I am trying to figure out how I would implement the devicons in the file.

kamiyaa:
I can probably provide a script later or something. The configuration for this I am still unsure of, so it might change in the future

gottaStartNow:
Scripting is good. Think about the configuration for a while then, ain't in no rush or nothing.

gottaStartNow:
Did you get around to this yet?

kamiyaa:
Sorry, I've been busy recently

gottaStartNow:
It is fine. Just love the project man.

kamiyaa:
Closing via #31

PR #31, opened 2020-10-08
Added basic implementation of devicons

SaeedAnas:
I have wrote a basic implementation for devicons support for file and directory names, which is requested in #3. Ported the dictionaries from [ranger devicons](https://github.com/alexanderjeurissen/ranger_devicons) into static hashmaps using the [rust phf crate](https://github.com/sfackler/rust-phf), and added configuration for showing the icons.


kamiyaa:
Hey, this is awesome! Great work! :heart: 

There is one thing I would like to point out so far.
Because the icon is part of the name now, sorting is not working as intended.
We will need to change the way `util/sort.rs` works so it doesn't work off of `direntry.file_name()` because that will give the name with the devicons :+1:

SaeedAnas:
To address the sorting bug, i've changed ```direntry.file_name()``` to ```direntry.raw_file_name()``` in ```sort::compare()```. Would any other modules also require this change, such as the renaming logic?

kamiyaa:
> Would any other modules also require this change, such as the renaming logic?

Yep! Good eye

SaeedAnas:
Instead of having to change all modules that need to use ```direntry.file_name()```, I thought it would be better to change the modules that will be rendering the filename, so i modified ```tui_dirlist``` and ```tui_dirlist_detailed``` to use ```direntry.label()```, which contains the icon and the filename. Would this be sufficient?

kamiyaa:
Looks good to me!

Issue #33, opened 2020-10-28
cd with tilde expansion


lahwaacz:
I'd like to configure a shortcut for `:cd ~` to quickly go to my home directory, but the command fails with

    No such file or directory (os error 2)

kamiyaa:
Apologies for not replying, I didn't notice this issue.

The current command line system is pretty scuffed.
Currently it doesn't do variable expansions and such.
This is mostly because it's complicated to get expansions to work while allowing for `:rename file name with space`

Ideally, there is a crate that does all this.
But I doubt there is one that allows for autocompletion on filenames with spaces.
So I predict there will be a lot of work to be done on the command line system (autocompletion, history, etc.)

kamiyaa:
closed via https://github.com/kamiyaa/joshuto/pull/38

PR #38, opened 2021-01-19
cd with tilde expansion

mblode:
#33 

I've added tilde expansion in the cd command. I also fixed an issue with the default mimetype config file.


kamiyaa:
Looks good overall. Any reason you chose to move to `dirs-next`?

mblode:
Awesome! `dirs-next` is just a fork of the abandoned `dirs` crate and `shellexpand` also uses it

kamiyaa:
I see. LGTM

PR #46, opened 2021-05-11
Add .webp support to default theme

nonetrix:
Now the default theme will highlight .webp with yellow just like .png


kamiyaa:
LGTM! Thanks!

PR #52, opened 2021-05-15
copying file path and directory to clipboard

makeefu:
Some simple ranger functionality concerning copying file names, file paths and current directories to clipboard added.

- New public functions `copy_filepath` and `copy_dir` which behave similarly to `copy_filename`.

- New `KeyCommand`s `CopyFilePath` and `CopyDir` behave like `CopyFileName`.

- Now one can copy file name, file path and current directory by calling `copy_filename`, `copy_filepath` and `copy_dir` commands.

- Respective shortcuts added to keymap.toml


src/commands/file_ops.rs
```diff
@@ -49,45 +49,92 @@ pub fn paste(context: &mut AppContext, options: IoWorkerOptions) -> JoshutoResul
...
+        .and_then(|entry| entry.file_path().to_str())
+        .and_then(|s| Some(s.to_string()));
+
+    if let Some(file_path) = entry_file_path {
+        copy_string_to_buffer(file_path)?;
+    }
+    Ok(())
+}
+
+pub fn copy_dir(context: &mut AppContext) -> JoshutoResult<()> {
```
kamiyaa, line 80: Could you change this to `copy_dirname` and same goes for `CopyDir` -> `CopyDirName`, just so its less ambiguous.

src/commands/file_ops.rs
```diff
@@ -49,45 +49,92 @@ pub fn paste(context: &mut AppContext, options: IoWorkerOptions) -> JoshutoResul
...
+    let opt_entry = context
+        .tab_context_ref()
+        .curr_tab_ref()
+        .curr_list_ref()
+        .and_then(|c| c.curr_entry_ref());
+
+    let dir:  Option<String> = opt_entry
```
kamiyaa, line 87: A bit confused on what this is trying to do:
let's say we have `/home/user/file.txt`, would this copy `/home/user` into the clipboard?
If so, you can use:
```rust
context
    .tab_context_ref()
    .curr_tab_ref()
    .curr_list_ref().file_path();
```

kamiyaa (changes requested):
Good work so far. I've requested some changes.
In addition, please run the code through `cargo fmt`

On another note, I will need to add CI/CD to dev branch.

src/commands/file_ops.rs
makeefu: Done!

src/commands/file_ops.rs
makeefu: Thank you! This is really simpler!

kamiyaa:
LGTM!

PR #56, opened 2021-05-17
Sorting

makeefu:
- `utils/sort.rs` refactored;
- sorting by file extension added;
- sort by extension key binding added;
- if you sort by extension, files with same extensions are sorted by previously enabled sorting method:

for example, if you consequently call
`sort lexical`
`sort ext`
you will get something like:

snse

or if you consequently call
`sort size`
`sort ext`
you will get something like:
ssse

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,0 @@
-/target
-/test
-*.log
-*.tmp
-*~
-**/*.rs.bk
--- a/config/keymap.toml
+++ b/config/keymap.toml
@@ -170,9 +170,6 @@ keys = [ "s", "n" ]
 [[mapcommand]]
 command = "sort reverse"
 keys = [ "s", "r" ]
-[[mapcommand]]
-command = "sort ext"
-keys = [ "s", "e" ]
 
 [[mapcommand]]
 command = "cd /"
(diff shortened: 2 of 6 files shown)


kamiyaa:
Very interesting! This is great work.

One thing I want to add is, does `pub fn mtime_sort` (in `src/util/sort.rs`) need to be changed as well?
I believe currently it defaults to `cmp::Ordering::Less` when the 2 entries are equal.

Other than that, I really appreciate the refactoring of the sorting code. It was pretty messy before.
Thanks!

kamiyaa:
Seems like something is broken. I'll merge once everything is fixed :+1:

makeefu:
Sorry, I just accidentally reverted last commit and don't know how to fix it. I'm not very experienced with git, lol.

kamiyaa:
> Sorry, I just accidentally reverted last commit and don't know how to fix it. I'm not very experienced with git, lol.

Haha no problem!
Some stuff you can look into later are `git rebase -i`, `git commit --amend`, `git push --force`.
Those are the commands I usually use when rewriting git history. :)

makeefu:
Thanks!
By the way I just found out that `sort_natural` and `sort_lexical` seem to be same, aren't they?

kamiyaa:
> Thanks! By the way I just found out that `sort_natural` and `sort_lexical` seem to be same, aren't they?

Oops forgot to reply to this.
`natural` and `lexical` should be different.

```shell
# lexical
IMG_000732.jpg
IMG_731.jpg

# natural
IMG_731.jpg
IMG_000732.jpg
```
Lexical compares character by character, whereas natural understand numbers

PR #58, opened 2021-05-20
Bookmarks

makeefu:
Ranger-like bookmarks:
Just call `add_bookmark` and type any char key to add the current directory to bookmarks.
Type `<apostrophe><char>`   to jump to a bookmarked directory.
The location of the file all the bookmarks are stored in is configurable.


src/config/default/config.rs
```diff
@@ -74,12 +78,15 @@ impl ConfigStructure for AppConfig {
 
 impl std::default::Default for AppConfig {
     fn default() -> Self {
+        let home = std::env::var("HOME").unwrap();
```
kamiyaa, line 81: These default values can be made into a function like `default_max_preview_size` and used here.
Then it can also be used with `#[serde(default = "func")]`.

Ideally, the default path should be where the rest of the config is; ie. `$XDG_CONFIG_HOME/joshuto/bookmarks.toml`

src/bookmarks.rs
```diff
@@ -0,0 +1,95 @@
+use crate::config::Flattenable;
```
kamiyaa, line 1: This should probably be placed under `src/config/default` or `src/config/bookmark`

src/context/app_context.rs
```diff
@@ -1,12 +1,14 @@
 use std::collections::VecDeque;
 use std::sync::mpsc;
 
+use crate::bookmarks::BookmarkMapping;
```
kamiyaa, line 4: I think I would probably just store the bookmarks inside `config::AppConfig`.
Any reason you decided to have it separate?

src/run.rs
```diff
@@ -58,10 +59,32 @@ pub fn run(
                             context.push_msg(e.to_string());
                         }
                     }
+                    Event::Key(Key::Char('`')) => {
```
kamiyaa, line 62: This should be delegated to its own command like `goto_bookmark` or something.
Then it can be mapped by users to any key.
You can take some inspiration from `open_with` and how joshuto currently handles multi-key command mappings.

src/ui/views/tui_bookmark_menu.rs
```diff
@@ -0,0 +1,127 @@
+use crate::context::AppContext;
```
kamiyaa, line 1: I didn't fully read and understand this file, but it seems to do something similar to `TuiMenu`?
I think it would be better to just create a widget that utilizes `TuiMenu` than to recreate the same logic.

src/bookmarks.rs
```diff
@@ -0,0 +1,95 @@
...
+
+            file.write_all(&s.as_bytes()).unwrap();
+        }
+        Ok(())
+    }
+}
+
+impl Flattenable<BookmarkMapping> for RawBookmarkMapping {
+    fn flatten(self) -> BookmarkMapping {
+        let mut bookmarksmap = BookmarkMapping::new();
```
kamiyaa, line 32: This can be reduced to
```rust
fn flatten(self) -> BookmarkMapping {
    let map: HashMap<char, PathBuf> = self.bookmarks
        .drain(..)
        .map(|(c, s)| (c, PathBuf::from(s))
        .collect();
    BookmarkMapping { map }
}
```

src/bookmarks.rs
```diff
@@ -0,0 +1,95 @@
...
+        }
+        return RawBookmarkMapping { bookmarks };
+    }
+
+    pub fn new() -> Self {
+        let map = HashMap::new();
+        Self { map }
+    }
+
+    pub fn load(bookmarks_file_path: &str) -> Self {
```
kamiyaa, line 65: This could return `std::io::Result<Self>`, `Result<Self, String>` or something and delegate the error handling to elsewhere.

kamiyaa:
Nice work overall!

Issue #59, opened 2021-05-21
Hidden files are traversing the column.


alannssantos:
# It happens only files, hidden folders behave correctly.

![Captura de tela_2021-05-21_18-46-44](https://user-images.githubusercontent.com/16065736/119201486-65f5c180-ba65-11eb-983c-5ef495b01e04.png)

kamiyaa:
Fixed via 7ef22c23c21db754a1321e7b23376615367527a3

PR #61, opened 2021-05-25
hidden files printing bug (issue 59)

makeefu:
The bug is that  hidden files which has ho extension are treated as extensions.
(Issue #59)

### DIFF
--- a/src/ui/widgets/tui_dirlist.rs
+++ b/src/ui/widgets/tui_dirlist.rs
@@ -116,7 +116,10 @@ fn print_entry(
                     } else {
                         (file_drawing_width - ext_width) as u16
                     };
-                    buf.set_string(x + ext_start_idx, y, extension, style);
+                    buf.set_stringn(x + ext_start_idx, y, extension,file_drawing_width, style);
+                    if ext_width > file_drawing_width {
+                        buf.set_string( x - 1 + file_drawing_width  as u16, y, ELLIPSIS, style);
+                    }
                     let ext_start_idx = if ext_start_idx > 0 {
                         ext_start_idx - 1
                     } else {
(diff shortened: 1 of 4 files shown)


src/ui/widgets/tui_dirlist.rs
```diff
@@ -90,9 +90,9 @@ fn print_entry(
         }
         _ => {
             let file_drawing_width = drawing_width;
-            let (stem, extension) = match name.rfind('.') {
-                None => (name, ""),
-                Some(i) => name.split_at(i),
+            let (stem, extension) = match entry.get_ext() {
```
kamiyaa, line 93: I think rather than changing `get_ext`'s behaviour, its better we add a case here instead.
Something like:
```rust
let (stem, extension) = match name.rfind('.') {
    None | Some(0) => (name, ""),
    Some(i) => name.split_at(name.len() - ext.len()),
};
```
Hopefully, this still works as intended xd

makeefu: Unfortunately this will not work because of devicons: `name` is actually the entry's label:
https://github.com/kamiyaa/joshuto/blob/f42660c2255a5db1a6c0918ac790bee7fae72d80/src/ui/widgets/tui_dirlist.rs#L80

makeefu:
I've just found out that the bug in issue #59 occur not when we deal with hidden files, but actually when we have too long extensions. I'll try to fix it.

kamiyaa:
Yea, no problem!

makeefu:
I don't quite understand what should `get_ext`'s behaviour be like?
For example, if a file's name is `.gitignore`, should it return `".gitignore"` or just `""`?

kamiyaa:
Fixed via 7ef22c23c21db754a1321e7b23376615367527a3

kamiyaa:
> I don't quite understand what should `get_ext`'s behaviour be like?
> For example, if a file's name is `.gitignore`, should it return `".gitignore"` or just `""`?

In my opinion, it should return `.gitignore`

PR #62, opened 2021-05-26
Clear the Screen Before Bulk Renaming

JMARyA:
A while after I do multiple bulk renames the screen becomes really messy containing files from previous bulk renames. This is a very simple fix. Maybe this could be improved a bit.

### DIFF
--- a/src/commands/bulk_rename.rs
+++ b/src/commands/bulk_rename.rs
@@ -81,6 +81,9 @@ pub fn _bulk_rename(context: &mut AppContext) -> JoshutoResult<()> {
         ));
     }
 
+    println!("{}", termion::clear::All);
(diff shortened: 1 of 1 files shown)


kamiyaa:
Could you provide a screenshot? I don't think I've ever encountered this. Thanks!

JMARyA:
Sure! On the left is the original joshuto and on the right my modified version doing two bulk renames. The version on the left side shows information on two bulk renames even though the first has already been done and is now irrelevant, whereas the right shows only the current bulk modifications making it easier to read with less distractions.
Bildschirmfoto 2021-05-29 um 02 45 56

kamiyaa:
Cool, if you could run the code through `cargo fmt`, I'll merge it

PR #66, opened 2021-06-06
Add a "touch" command

DLFW:
This PR adds a "touch" command to joshuto. 

If invoked without an argument, the currently selected file’s or directory’s atime and ctime are updated to the current system time.

If invoked with an argument that matches an existing file or dir to the, that file’s or dir’s atime and ctime is updated.

If invoked with an argument that does not match an existing file or dir, a new empty file with that name is created.

This PR introduces a dependency to the "filetime" crate.
I'm very new to Rust and I hope that there are no severe flaws in here. :)
I don’t fully get the "joshuto logic" parts for updating the view. (Some code documentation would be great ;) ) I tried to get that stuff fine by taking snippets from other commands.

Unfortunately, I merged from upstream in between which now makes squashing a bit problematic. If the commits are too fine grained, I can recreate the feature on a clean branch.


Cargo.toml
```diff
@@ -12,6 +12,7 @@ alphanumeric-sort = "^1"
 chrono = "^0"
 colors-transform = "^0"
 dirs-next = "^2"
+filetime = "0.2"
```
kamiyaa, line 15: Please change this to "^0", just my preference xd

src/commands/touch_file.rs
```diff
@@ -0,0 +1,48 @@
...
+        .curr_tab_ref()
+        .curr_list_ref()
+        .and_then(|s| s.curr_entry_ref())
+        .map(|s| s.file_path().to_path_buf());
+
+    if let Some(selected_file_path) = selected_file_path {
+        match arg {
+            "" => _update_actime(&selected_file_path)?,
+            file_arg => {
+                let mut file = context.tab_context_ref().curr_tab_ref().cwd().to_path_buf();
```
kamiyaa, line 32: This isn't really necessary because the `context.tab_context_ref().curr_tab_ref().cwd()` should equal to the current directory.
So your file could just be:
```rust
let file = path::PathBuf::from(file_arg);
```

kamiyaa:
In addition to the changes I requested, I just want to know what the use case for this is?

There is the `shell` command that allow users to run shell commands (albeit very rudimentary and definitely needs work)

DLFW:
Hi @kamiyaa, thanks for the feedback.
Yes, I saw the shell command and I had the same thought. My motivation:

1. How can I pass the currently selected file to it in an easy way? Is there some kind of variable set available for shell based commands which can be used to "paste" the currently selected file or the marked files? (BTW, would be extremely useful in general, for example to open files with another file-type-independent program, like a dmenu based "open with" dialog, based on rifle. ;) )
2. Also a solution with "variables" would be less convenient for users when they just want to touch the selected file, unless they configure two separate key bindings for "touch selected" and "touch specified/non-existent" file (I guess).
3. When joshuto allows "auto-completion" (like ranger with TAB), the command can be ... well,  auto-completed, simply because it's inbuilt.
4. People coming from ranger may expect it. (Like I did.)

Not really an argument but a related question: What is the reason for having the `mkdir` command implemented? That could also be achieved by a shell call. I think there is some benefit in having the most basic file operations implemented. But to be honest, I also took this as a simple exercise for programming Rust. If you think this is too bloated and want to keep this feature out of joshuto, I'm also ok with that.

BTW: Thanks for the great project!

kamiyaa:
> 1. How can I pass the currently selected file to it in an easy way? Is there some kind of variable set available for shell based commands which can be used to "paste" the currently selected file or the marked files? (BTW, would be extremely useful in general, for example to open files with another file-type-independent program, like a dmenu based "open with" dialog, based on rifle. ;) )

Yeah, this would be nice. I'm mostly too lazy to write parsing code to figure out replacing `$f` or something to the selected/marked files xd

Issue #69, opened 2021-06-10
Spawn commands executed with ":shell"


DLFW:
When using a command with :shell, like 
```toml
[[mapcommand]]
command = "shell sxiv -t %s &>/dev/null &"
keys = [ "i" ]
```
joshuto waits for the process to terminate at `command.status()?;` in `shell.rs`.

For many use cases, this is pretty unhandy. Instead of using a blocking sub-process call, joshuto could just spawn the sub-process with `command.spawn()?;` instead.

Alternatively, joshuto could use two shell commands, one blocking and one non-blocking. (I could take that.)

kamiyaa:
Yeah, the problem with spawning it in the background is the output of the program will mess up joshuto's UI.

I've thought of coming up with a way to let it run in the background by having a `--fork` option or something, but it would be hard to distinguish between `--fork` and a shell argument.

I think we can have two commands:
1. `shell`: blocking
2. `spawn`: non-blocking

DLFW:
I like it. Of course it would be good to prevent any command from messing up the UI by stdout/err in general.
But providing such two commands is...

1.  an acceptable workaround and...
2.  maybe a useful differentiation in general. Maybe joshuto wants to use the output of blocking commands later on for other things like showing error messages or taking decisions on updating the cwd content.

I definitely prefer a two-command solution over a solution with a `--fork`-like option which would at least be confusing for the user. 

In Ranger, an ampersand at the end of the command does the job. For some reason this doesn't work in Rust's `std::process::Command`. Maybe using two different commands is even better for some portability stuff or something in the future. 

Maybe I'll give it a try if I have some time this evening.

DLFW:
I just provided a PR for this and I need to correct myself in two aspects:

1. Swallowing the stdout/stderr for the blocking version (`shell`) might not be a good idea because I realized that I sometimes `Ctrl-Z` ranger to the background to inspect the output of a command and the `fg` ranger back to foreground. It's necessary in that case to let the sub-process print to the terminal. BTW, joshuto catches `Ctrl-Z`. It doesn't reach my terminal, so no TSTP signal is send. I might create another Issue for that.
2. I just realized that Rust's `std::process::Command` simply doesn't execute a shell around the process to start. (Not really a surprise.) I guess the user can just start a shell explicitly. (Haven't tried...) That's even better since users can choose their shell and portability issues for sub-processes are more in the hand of the users. However, I find the command name `shell` a bit misleading in this case.

[...]

PR #70, opened 2021-06-11
Add spawn command

DLFW:
As discussed in  #69, this PR adds a `spawn` command.

When executing a command with spawn, stdout and stderr of the executed sub-process are swallowed, so they don't mess up Joshuto's UI.


kamiyaa:
LGTM! Thanks!

Issue #71, opened 2021-06-12
Rework file entry representation in “TuiDirListDetailed”


DLFW:
Unlike ranger, joshuto does not show the size of directories (number of files and dirs inside), and does not show any size information for symlinks. Also, joshuto shows symlinked directories like files and with a wrong devicon.

Two views on the same test directory:

Ranger:
![ranger_file_view](https://user-images.githubusercontent.com/22507197/121788867-59cad300-cbd1-11eb-8424-1b37f198bd5d.png)

Joshuto:
![joshuto_file_view](https://user-images.githubusercontent.com/22507197/121788896-972f6080-cbd1-11eb-82d5-7e48602d6eae.png)

I would propose to rework joshuto's display to match ranger's. It just makes sense and the "size" of directories is quite useful.

Furthermore, ranger supports multiple (and custom) views (called “linemode”), which I use for example to show [tmsu](https://tmsu.org/) tags for each file and dir. But that's likely worth its own issue.

---
Edit: One more difference where the ranger variant makes more sense: When a symlink is selected, the footer bar shows the permission flags of the linked target rather those of the symlink itself, which is always `lrwxrwxrwx`.

I have started an experimental implementation.

kamiyaa:
👍 
Yes, this was something I was aware of, but haven't really found it that annoying so I never got around to it xd.

I believe how ranger does it is, it only shows the number of files/directories inside a folder if and only if the cursor has been on that folder before. I'm guessing this is because previously, the contents of the folder has not been loaded into their data structures.

I'm guessing your implementation would have similar behaviour? Where if the contents of the folder is not loaded in our HashMap, its not displayed.

> One more difference where the ranger variant makes more sense: When a symlink is selected, the footer bar shows the permission flags of the linked target rather those of the symlink itself, which is always lrwxrwxrwx.

I never knew this was a thing. Learning something new everyday :)

> Furthermore, ranger supports multiple (and custom) views (called “linemode”), which I use for example to show tmsu tags for each file and dir. But that's likely worth its own issue.

Yea, if you could elaborate this on a new issue, that'd be great! I'm not very familiar with ranger's different views.

DLFW:
> I believe how ranger does it is, it only shows the number of files/directories inside a folder if and only if the cursor has been on that folder before.

Nope. It shows them always right from the beginning.

---

I just recognize that there's a third feature that would fit in this context:

[...]

Issue #29, opened 2020-07-30
Enable Github Actions for Repository.


UsairimIsani:
Please enable Github Actions for the CI to take effect.

kamiyaa:
Closed via 8c9500478c09fe8a224309605e12c5f52f32db62

There seemed to be an issue with the action not being detected. Went online and found that I had to delete and revert

PR #72, opened 2021-06-15
Show symlinked files and dirs correctly

DLFW:
This PR is the first part of #71. It does not contain the directory size feature, but it includes:

* Show symlinked dirs like dirs, with a directory devicon
* Show size of target file for symlinked files in conbination with the “symlink arrow”
* Show permission flags of target file for symlinked files

This PR is much more invasive than I expected. Sorry for that. But when I started to combine the “symlink arrow” with the file size, one change led to the next. 

The massive change is mostly because of the calculation of the entry strings to display, which is now done in separate functions without using the `buf.set_stringn` method to handle the truncation of strings.

The good thing is that the code is more testable and easier to extend, for my feeling. A downside might be that the string truncation is based on characters. I dont't know how tui-rs handles that, but I guess it uses correct graphemes instead of just characters when using the `buf.set_stringn` function. I could use something like [unicode-segmentation](https://crates.io/crates/unicode-segmentation/1.7.1) instead to realize a more proper truncation.

Also, some refactoring for `pub fn entry_style` would be good, it really looks kind of ugly, I know.

Anyway, before putting more effort into this, I wanted to give a chance for feedback.

### DIFF
--- a/src/fs/metadata.rs
+++ b/src/fs/metadata.rs
@@ -3,8 +3,13 @@ use std::{fs, io, path, time};
 #[derive(Clone, Debug)]
 pub enum FileType {
     Directory,
-    Symlink(String),
     File,
+}
+
+#[derive(Clone, Debug)]
+pub enum LinkType {
+    Normal,
+    Symlink(String),
 }
 
 #[derive(Clone, Debug)]
@@ -13,6 +18,7 @@ pub struct JoshutoMetadata {
     _modified: time::SystemTime,
     _permissions: fs::Permissions,
     _file_type: FileType,
+    _link_type: LinkType,
     #[cfg(unix)]
     pub uid: u32,
     #[cfg(unix)]
@@ -26,40 +32,43 @@ impl JoshutoMetadata {
         #[cfg(unix)]
         use std::os::unix::fs::MetadataExt;
 
-        let metadata = fs::symlink_metadata(path)?;
+        let symlink_metadata = fs::symlink_metadata(path)?;
+        let metadata = fs::metadata(path)?;
 
         let _len = metadata.len();
         let _modified = metadata.modified()?;
         let _permissions = metadata.permissions();
-        let file_type = metadata.file_type();
+        let _file_type = match metadata.file_type().is_dir() {
+            true => FileType::Directory,
+            false => FileType::File,
+        };
+        let _link_type = match symlink_metadata.file_type().is_symlink() {
+            true => {
+                let mut link = "".to_string();
 
-        let file_type = if file_type.is_dir() {
-            FileType::Directory
-        } else if file_type.is_symlink() {
-            let mut link = "".to_string();
-
-            if let Ok(path) = fs::read_link(path) {
-                if let Some(s) = path.to_str() {
-                    link = s.to_string();
+                if let Ok(path) = fs::read_link(path) {
+                    if let Some(s) = path.to_str() {
+                        link = s.to_string();
+                    }
                 }
+                LinkType::Symlink(link)
             }
-            FileType::Symlink(link)
-        } else {
-            FileType::File
+            false => LinkType::Normal,
         };
 
         #[cfg(unix)]
-        let uid = metadata.uid();
+        let uid = symlink_metadata.uid();
         #[cfg(unix)]
-        let gid = metadata.gid();
+        let gid = symlink_metadata.gid();
         #[cfg(unix)]
-        let mode = metadata.mode();
+        let mode = symlink_metadata.mode();
 
         Ok(Self {
             _len,
             _modified,
             _permissions,
-            _file_type: file_type,
+            _file_type,
+            _link_type,
             #[cfg(unix)]
             uid,
             #[cfg(unix)]
@@ -87,5 +96,9 @@ impl JoshutoMetadata {
 
     pub fn file_type(&self) -> &FileType {
         &self._file_type
+    }
+
+    pub fn link_type(&self) -> &LinkType {
+        &self._link_type
     }
 }
--- a/src/fs/metadata.rs
+++ b/src/fs/metadata.rs
@@ -38,9 +38,10 @@ impl JoshutoMetadata {
         let _len = metadata.len();
         let _modified = metadata.modified()?;
         let _permissions = metadata.permissions();
-        let _file_type = match metadata.file_type().is_dir() {
-            true => FileType::Directory,
-            false => FileType::File,
+        let _file_type = if metadata.file_type().is_dir() {
+            FileType::Directory
+        } else {
+            FileType::File
         };
         let _link_type = match symlink_metadata.file_type().is_symlink() {
             true => {
--- a/src/ui/widgets/tui_dirlist_detailed.rs
+++ b/src/ui/widgets/tui_dirlist_detailed.rs
@@ -219,58 +219,57 @@ mod test_factor_labels {
             )
         );
     }
+}
+
+#[cfg(test)]
+mod test_trim_file_label {
+    use super::trim_file_label;
 
     #[test]
     fn dotfiles_get_an_ellipsis_at_the_end_if_they_dont_fit() {
-        let left = ".joshuto";
-        assert_eq!(
-            (".jos…".to_string(), "".to_string()),
-            factor_labels_for_entry(left, "".to_string(), 5)
-        );
+        let label = ".joshuto";
+        assert_eq!(".jos…".to_string(), trim_file_label(label, 5));
     }
 
     #[test]
     fn dotless_files_get_an_ellipsis_at_the_end_if_they_dont_fit() {
-        let left = "Desktop";
-        assert_eq!(
-            ("Desk…".to_string(), "".to_string()),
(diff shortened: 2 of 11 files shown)


src/ui/widgets/tui_dirlist_detailed.rs
```diff
@@ -80,113 +80,197 @@ fn print_entry(
...
-                return;
-            }
-            let file_drawing_width = drawing_width - SYMLINK_WIDTH;
-
-            print_file_name(buf, (x, y), name, style, file_drawing_width);
+    let size_string = match entry.metadata.file_type() {
+        FileType::Directory => String::from(""),
+        FileType::File => format::file_size_to_string(entry.metadata.len()),
+    };
+    let symlink_string = match entry.metadata.link_type() {
```
kamiyaa, line 87: This part probably doesn't need to be a String, and can just be a static str, unless you're planning to do more with it later.
```rust
    let symlink_string = match entry.metadata.link_type() {
        LinkType::Normal => "",
        LinkType::Symlink(_) => "-> ",
    };
```

src/fs/metadata.rs
```diff
@@ -26,40 +32,43 @@ impl JoshutoMetadata {
...
-
-        let file_type = if file_type.is_dir() {
-            FileType::Directory
-        } else if file_type.is_symlink() {
-            let mut link = "".to_string();
+        let _file_type = match metadata.file_type().is_dir() {
+            true => FileType::Directory,
+            false => FileType::File,
+        };
+        let _link_type = match symlink_metadata.file_type().is_symlink() {
```
kamiyaa, line 45: These true/false matches aren't really necessary and can just be if/else

src/ui/widgets/tui_dirlist_detailed.rs
```diff
@@ -80,113 +80,197 @@ fn print_entry(
...
-                };
-                buf.set_string(x + ext_start_idx, y, ELLIPSIS, style);
-            }
+            let stem_width = drawing_width - ext_width;
+            let truncated_stem = truncate_by_char(&stem.to_string(), stem_width - 1);
+            format!("{}{}{}", truncated_stem, ELLIPSIS, extension)
         }
     }
 }
+
```
kamiyaa, line 162: I would prefer if the tests were in their own separate file. In this case, something like, `src/ui/widgets/tui_dirlist_detailed_test.rs` or something

DLFW:
Hi and thanks, @kamiyaa !
I think we need to go with option 1, because of two reasons:
* if we don't do truncation on our own, we would loose test-ability again and “label shortening” would always depend on the TUI-context (also it just feels like an unnecessary deep dependency)
* as soon as we want to support more complex label arrangements (for other “linemodes”), we will need to have a function for this anyway

Also, `set_stringn` will have to  do the same calculation with the same complexity internally. We don't win anything when using tui for this.

Luckily, with [unicode-segmentation](https://crates.io/crates/unicode-segmentation/1.7.1) we should be able to solve it like this:
```rust
#[inline(always)]
pub fn truncate_by_grapheme(s: &String, max_chars: usize) -> String {
    let graphemes: Vec<&str> = s.graphemes(true).collect();
    if graphemes.len() <= max_chars {
        String::from(s)
    } else {
        graphemes[0..max_chars].concat()
    }
}
```
`unicode-segmentation` is even a dependency of tui-rs, so it's already in and tui will likely use the same algorithms internally.

I will give it a try in the next days.

kamiyaa:
Sounds good to me!

kamiyaa:
We can also look into potentially remove unicode-width dependency if unicode-segmentation covers the same use cases

DLFW:
As far as I can tell, it does not offer such a nice integration or even a ready-to-use function, but it would be possible to write one like for the truncation.

I'm a bit confused with all these Unicode rules, though. `unicode-width` says it “determines displayed width of char and str types according to [Unicode Standard Annex #11 rules](http://www.unicode.org/reports/tr11/)”. `unicode-segmentation` says it “splits strings on Grapheme Cluster or Word boundaries, according to the [Unicode Standard Annex #29 rules](http://www.unicode.org/reports/tr29/)”. Does a grapheme always correspond to one cell in a terminal? 

I would propose we just give it a try and test it, unless some Unicode nerd comes around and explains us this stuff... :)

kamiyaa:
> I would propose we just give it a try and test it, unless some Unicode nerd comes around and explains us this stuff... :)

Sure, let me know when your changes are ready to be merged. 👍

src/ui/widgets/tui_dirlist_detailed.rs
DLFW: The [rust book says](https://doc.rust-lang.org/stable/book/ch11-03-test-organization.html#unit-tests) “The convention is to create a module named tests in each file to contain the test functions and to annotate the module with cfg(test).”

When first checking how to do unit tests in Rust, I found this also a bit... special. However, after some thoughts it somehow makes sense and it seems to be the Rust convention.

Anyway, if you want, I can move them. But maybe spend a second though on this first. ;)

kamiyaa: Yea, we can have a test module instead :)

DLFW:
Hi @kamiyaa, can you check the file name display again with “some files containing lots of asian characters”?
I tested it with some  random copy-paste strings from various Wikipedias and with some unit tests, and I have some hope that it works now.

BTW: I saw that tui-rs depends also on `unicode-width`. I checked [their truncation algorithm](https://github.com/fdehau/tui-rs/blob/fbd834469fd99c7408a20ed09c5e205fe994c678/src/buffer.rs#L293) and I think I understood a little more about the problem. Seems there are characters with 0 (diacritics), 1 (“normal”), and 2 (“full-width”) columns, and on top of that, not on all positions, a string-truncation makes sense (e.g. within graphemes). So, both libs (`unicode-width` and `unicode-segmentation`) are needed, one to check the columns taken by a certain string, and one to check where the string can be split.

kamiyaa:
This is looking much better!
The filenames no longer write over their specified width.

Just one minor edge case:
When the last character takes up 2 columns, the ... does not get drawn.

Though, I don't think this is a major problem. The current joshuto has the same problem.
So if you're content with the changes, I can merge it :)
image

DLFW:
Hmmmm.... so the last character shown is not the last character in the file/dir name?
Can you paste a file name here for me to reproduce the problem?
(What script is that, by the way? :) )

kamiyaa:
> Hmmmm.... so the last character shown is not the last character in the file/dir name?
> Can you paste a file name here for me to reproduce the problem?

Yeah sure.
The file names I have are: (2 directories, but files with extensions have the same problem)
```
一二三四五六▌一二三四五▌一二三四五六
一二三四五六一二三四五一二三四五六一二三四五六
```

> (What script is that, by the way? :) )

script?

DLFW:
I mean “writing system” / alphabet. 
At first, it was just out of curiosity; now it seems I need to know that because neigther my browser not my terminal show any useful characters. I might need to choose an appropriate font, I guess.

kamiyaa:
Ohh I see. It is Chinese characters. You can probably get it to work by installing noto fonts from google and setting your terminal font to use that

DLFW:
Hm, can't get it work right now. Anyway, just wait with the merge. The unit test also need to go into their own modules still. ;)

src/ui/widgets/tui_dirlist_detailed.rs
DLFW: Hm, so that means that the test modules must be listed in the `mod.rs` and that also means that we are **forced to make private functions public, just to write unit tests**, even if it doesn't make sense to have them declared public. Are you really, really sure? What's the reason behind this? (I haven't thought about this privacy-problem myself before...)

I had just started to rewrite that, but I'm really feeling like doing something very wrong here. Making functions public which should be hidden in their module gives me a severe stomach ache...

kamiyaa: Oh I see. I didn't account for private functions and stuff.

Okay, we can leave it as it is since its the convention stated by the Rust book.

DLFW:
I cannot reproduce the fullwidth-ellipsis error. No matter how I resize the terminal, it always looks ok:
chinese_characters_correct
I also have added [two more unit tests](https://github.com/DLFW/joshuto/blob/2a36a0e84dc7b4b62215e26973ddffa495a7027d/src/ui/widgets/tui_dirlist_detailed.rs#L264), hoping to be able to reproduce it, but they pass....

From my side, this is ready for merge now. :)

DLFW:
Hi @kamiyaa, can you merge this one? Maybe I can continue then soon with the directory-size feature.

kamiyaa:
> Hi @kamiyaa, can you merge this one? Maybe I can continue then soon with the directory-size feature.

Sure!

> I cannot reproduce the fullwidth-ellipsis error. No matter how I resize the terminal, it always looks ok:

Hmmm, maybe its my terminal then.

> Edit: BTW: There any quite some warnings on build. Seems they are related to the preview-changes. Have you merged dev to main?

Yes, I made a few changes to previewing support. Nothing drastic yet. Will need a lot more time to sit down and plan things out. :/

PR #75, opened 2021-06-19
open command line with a colon

DLFW:
Beside opening the command line with a semi-colon, it can be opened also
with a colon. The default key mapping has been adapted as well as the
keymap.toml configuration template.


kamiyaa:
LGTM!

PR #76, opened 2021-06-20
show size of directories

DLFW:
The middle column shows the size of directories in terms of contained
files and sub-directories in the right label for each entry.

Invalid symlinks and hidden files/dirs are included in the count.

### DIFF
--- a/src/fs/metadata.rs
+++ b/src/fs/metadata.rs
@@ -15,6 +15,7 @@ pub enum LinkType {
 #[derive(Clone, Debug)]
 pub struct JoshutoMetadata {
     _len: u64,
+    _directory_size: Option<usize>,
     _modified: time::SystemTime,
     _permissions: fs::Permissions,
     _file_type: FileType,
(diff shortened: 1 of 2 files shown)


DLFW:
Second feature for #71.

There is one design decision that could be discussed: The size of directories is hold by an `Option<usize>`:
```rust
pub struct JoshutoMetadata {
    _len: u64,
    _directory_size: Option<usize>,
    // ...
}
```
It could have been a simple `usize` also, or I could have even reused the existing `_len` attribute. I chose this option because “explicit is better than implicit”. :wink: 

To keep the code more consistent, `_len` could be changed and renamed to `_file_size: Option<u64>`.
Opinions?

I`m also fine with just merging it.

kamiyaa:
> Second feature for #71.
> 
> There is one design decision that could be discussed: The size of directories is hold by an `Option<usize>`:
> 
> ```rust
> pub struct JoshutoMetadata {
>     _len: u64,
>     _directory_size: Option<usize>,
>     // ...
> }
> ```
> 
> It could have been a simple `usize` also, or I could have even reused the existing `_len` attribute. I chose this option because “explicit is better than implicit”. wink
> 
> To keep the code more consistent, `_len` could be changed and renamed to `_file_size: Option<u64>`.
> Opinions?
> 
> I`m also fine with just merging it.

I think being explicit is good :+1:

kamiyaa:
I've made a few changes to the way preview works.
Mainly, the previewing logic is centralized in `run.rs` now instead of scattered across different commands. Just giving a heads up in case it conflicts with some of your changes or your planned changes. (58ad2824f631dcf8ff2a3f267733ffc0c302b3b3)

DLFW:
Thanks!

But if you're OK with keeping the `_len` attribute as it is for now, I'm perfectly fine with merging this one (if it still doesn't conflict).

kamiyaa:
I think `_len` is good :+1:

kamiyaa:
Made a minor fix to make sure directories without proper permissions (like `/root`) are still displayed.

And added a one space padding just like in ranger (72aaf0c)

PR #77, opened 2021-06-20
Fix selecting an entry by mouse click when show_borders=true

DLFW:
This PR fixes two bugs related to selecting entries with a left-mouse-click which appeared when borders are enabled.

It also adds a tiny code enhancement by putting the calculation of the “paging” into one central place, where it can be changed/enhanced/maintained more easily.

### DIFF
--- a/src/util/input.rs
+++ b/src/util/input.rs
@@ -16,7 +16,13 @@ pub fn process_mouse(event: MouseEvent, context: &mut AppContext, backend: &mut 
     let constraints: &[Constraint; 3] = &context.config_ref().display_options_ref().default_layout;
     let layout_rect = Layout::default()
         .direction(Direction::Horizontal)
-        .vertical_margin(1)
+        .vertical_margin(
+            if context.config_ref().display_options_ref().show_borders() {
+                2
+            } else {
+                1
+            },
+        )
         .constraints(constraints.as_ref())
         .split(f_size);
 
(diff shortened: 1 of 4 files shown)


src/util/input.rs
```diff
@@ -54,21 +54,27 @@ pub fn process_mouse(event: MouseEvent, context: &mut AppContext, backend: &mut
         MouseEvent::Press(MouseButton::Left, x, y)
             if y > layout_rect[1].y && y <= layout_rect[1].y + layout_rect[1].height =>
         {
-            if x < layout_rect[1].x {
-                if let Some(dirlist) = context.tab_context_ref().curr_tab_ref().curr_list_ref() {
```
DLFW, line 58: The actual bug was in this line by using `curr_list_ref` instead of `parent_list_ref`.

src/util/input.rs
```diff
@@ -16,7 +16,13 @@ pub fn process_mouse(event: MouseEvent, context: &mut AppContext, backend: &mut
     let constraints: &[Constraint; 3] = &context.config_ref().display_options_ref().default_layout;
     let layout_rect = Layout::default()
         .direction(Direction::Horizontal)
-        .vertical_margin(1)
```
DLFW, line 19: This fixed the border-bug by providing the correct view port height and the correct y-origin for the entry lists.

kamiyaa:
LGTM! Thanks for the great work!

PR #78, opened 2021-06-21
fix file sizes not appearing when filename too long

kamiyaa:
@DLFW Just wanted to touch base with you on this change.

I made a few changes to your rendering process.
The minor part is just reducing the amount of memory allocations by using lifetimes.
The more major part is fixing (I think this is unintentional behavior?) file sizes not showing up when the file name is long enough, but not too long to cause `...` to appear.

Expected:
20210620-205334_swappy

Actual:
20210620-205341_swappy


DLFW:
Hi @kamiyaa, if I got you right, it's not a bug, it's a feature. :smile: 

If we don't have enough space to show both, the size and a feasible amount of characters from the file name, the current implementation gives precedence to the file name, since it's the more important information.

You can find this behavior described in the unit tests [`right_label_omitted_if_left_label_would_need_to_be_shortened_below_min_left_label_width`](https://github.com/kamiyaa/joshuto/blob/cf660136ea43e0a463fd270bfdaf8bb3ee606d9d/src/ui/widgets/tui_dirlist_detailed.rs#L198) and [`right_label_is_kept_if_left_label_is_not_shortened_below_min_left_label_width`](https://github.com/kamiyaa/joshuto/blob/cf660136ea43e0a463fd270bfdaf8bb3ee606d9d/src/ui/widgets/tui_dirlist_detailed.rs#L209).

The "feasible amount of characters" is defined by `MIN_LEFT_LABEL_WIDTH`, which is currently set to 15. I guess a value of 8 would maybe make more sense. Just reducing that constant's value should give you the desired behavior. So, if it's set to 8, the right label will be omitted if a file name would need to be shortened to a length of 8 or less. The unit tests may need adaptation when changing the value.

kamiyaa:
Ohh I see. However, that's not the behaviour I am encountering.

20210621-142301_swappy

In this example, the file name is clearly longer than 15 columns, but the file size is not shown.

I think its because of this line:
```rust
if left_label_original.width() <= drawing_width {
```
when it should be
```rust
if left_label_original.width() + right_label_original.width() <= drawing_width {
```

DLFW:
Oh, ok, but
```rust
left_label_original_width + right_label_original_width <= drawing_width
```
is equal to the 
```rust
let left_width_remainder = drawing_width as i32 - right_label_original_width as i32;
    let width_remainder = left_width_remainder as i32 - left_label_original_width as i32;
    if width_remainder >= 0 {
```
before.... am I too tired?

kamiyaa:
Oh you're right. Maybe we just remove this if statement entirely then :thinking:

DLFW:
Yeah, the bug is there, but if we just omit it, an ellipsis is printed on the last character:
```
  left: `("foobarbazfo.ext", "")`,
 right: `("foobarbazf….ext", "")
```
But that should be easy to fix... I'll have a look

DLFW:
I think this exemplary change would fix it without introducing other problems: https://github.com/DLFW/joshuto/commit/608b7b6f14aa5279e4870763bdaa1136434712f4

I can also provide a cleaner version based on `dev` later.

kamiyaa:
Sure, thanks!

I'll merge this and you can work ontop of it later.

DLFW:
created #79 to solve this

PR #80, opened 2021-06-27
highlight invalid symlinks

### LINKED ISSUE #71: Rework file entry representation in “TuiDirListDetailed”

DLFW:
Dangling symlinks are shown with a separate style. `theme.toml` has been extended with a `[link_invalid]` section. The default style is red+bold.

Furthermore, the link target, which is shown in the footer for a selected symlink, uses either the “link” style or the ”link_invalid” style respectively.

Merge request for `dev` branch as I saw that a symlink related commit was already on that branch. 
We can close #71 then. :wink:

### DIFF
--- a/config/theme.toml
+++ b/config/theme.toml
@@ -20,6 +20,10 @@ bold = true
 fg = "light_cyan"
 bold = true
 
+[link_invalid]
+fg = "red"
+bold = true
+
 [socket]
 fg = "light_magenta"
 bold = true
--- a/src/config/theme/app_theme.rs
+++ b/src/config/theme/app_theme.rs
@@ -19,6 +19,8 @@ pub struct RawAppTheme {
     #[serde(default)]
     pub link: RawAppStyle,
     #[serde(default)]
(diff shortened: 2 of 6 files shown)


kamiyaa:
LGTM! Thanks!

PR #83, opened 2021-07-12
Add copying filename without extension

krivahtoo:


src/commands/file_ops.rs
```diff
@@ -62,6 +62,20 @@ pub fn copy_filename(context: &mut AppContext) -> JoshutoResult<()> {
     Ok(())
 }
 
+pub fn copy_filename_without_extension(context: &mut AppContext) -> JoshutoResult<()> {
+    let entry_file_name = context
+        .tab_context_ref()
+        .curr_tab_ref()
+        .curr_list_ref()
+        .and_then(|c| c.curr_entry_ref())
+        .map(|entry| entry.file_name().rsplit_once(".").unwrap().0.to_string());
```
kamiyaa, line 71: Will this crash if the file name does not contain an extension?

krivahtoo: Yes, I did not try that earlier.
I'm working on a fix.

kamiyaa:
LGTM! Thanks!

PR #87, opened 2021-07-23
Add tab switching by index

krivahtoo:
Add tab switching by index for ranger-like key bindings `Alt+1,Alt+2,..`


src/commands/key_command.rs
```diff
@@ -345,6 +347,13 @@ impl std::str::FromStr for KeyCommand {
                     format!("{}: {}", command, e.to_string()),
                 )),
             },
+            "tab_switch_index" => match arg.parse::<i32>() {
```
kamiyaa, line 350: This can be changed to `usize` or `u32` since index is always going to be positive

kamiyaa:
LGTM, thanks!

PR #95, opened 2021-09-28
Add a help page

### LINKED ISSUE #84: Show Keybindings when pressing <?>
Most TUI programs have a useful list of categorized keybindings that shows when `<?>` or `shift`+`/` are pressed to quickly check which-key does X, or see what commands are available to the user. I believe this is a low-input/high-return improvement for all new users. This would also make the project feel much more polished.

Vinegret43:
closes #84

Add a help page with keybindings, shown when pressing ```?```

### Features:
- Scrolling with j and k, arrows, PageUp, PageDown, Home, End or mouse wheel
- Coloring
- Short description for every command
- Allowing user to choose by which field to sort (With 1, 2, 3)
- Search


kamiyaa:
This is really great work @Vinegret43 !

However, I'm not too excited about the big changes to `run.rs` main event loop.

I understand the reasoning because `keymap_t` is not readily available to commands.
But I think I would rather change `KeyCommand.execute()` to take in an immutable reference to `keymap_t`
than to modify the existing event loop.

Kind of like how its done currently with `src/commands/show_workers.rs`.
Though most of the logic for `show_workers.rs` is currently in `src/ui/views/tui_worker_view.rs`.
(I think its definitely possible to move that logic out of the view and into `show_workers.rs`)
(I will probably do that at some point)

Thoughts would be appreciated!

Also please run your code through `cargo fmt`, otherwise I can't (and won't) merge it :stuck_out_tongue:

inoshy:
all checks failed. alas.

Vinegret43:
Thanks for your response, it's definitely a better idea to make the loop inside ```KeyCommand.execute()``` instead of changing the main loop in ```run.rs```, i just didn't really think of that. Also, next time i'll make sure i run ```cargo fmt``` before commiting

Vinegret43:
Also, what's the reason for separating ```AppContext``` and ```AppKeyMapping```? Why can't we just put keymapping right into context?

kamiyaa:
> Also, what's the reason for separating `AppContext` and `AppKeyMapping`? Why can't we just put keymapping right into context?

I think the main reason was because inside `run.rs` I do a `keymap_t.get()`, but then `AppContext` needs to be mutable when passed into `execute()`.
There's probably a way around this, which will allow us to put `keymap_t` inside `AppContext`.
Something like:
```rust

let key = keymap_t.get(input).clone(); // this prevents a immutable borrow

if let Some(command) = key {
    command.execute(context, backend);  // mutable borrow
}
```
Personally, I'm not really into cloning stuff around (though I have to do it myself in various places xd)

inoshy:
> > Also, what's the reason for separating `AppContext` and `AppKeyMapping`? Why can't we just put keymapping right into context?
> 
> I think the main reason was because inside `run.rs` I do a `keymap_t.get()`, but then `AppContext` needs to be mutable when passed into `execute()`. There's probably a way around this, which will allow us to put `keymap_t` inside `AppContext`. Something like:
> 
> ```rust
> let key = keymap_t.get(input).clone(); // this prevents a immutable borrow
> 
> if let Some(command) = key {
>     command.execute(context, backend);  // mutable borrow
> }
> ```
> 
> Personally, I'm not really into cloning stuff around (though I have to do it myself in various places xd)

congrats on half a thousand stars!

kamiyaa:
Thanks @inoshy !

inoshy:
you are welcome! I got hooked up with your joshuto :) love it.

src/commands/help.rs
```diff
@@ -0,0 +1,90 @@
...
+        let keymap = if search_query.is_empty() {
+            widgets::get_keymap_table(keymap_t, &search_query, sort_by)
+        } else {
+            widgets::get_keymap_table(keymap_t, &search_query[1..], sort_by)
+        };
+        backend.render(TuiHelp::new(&keymap, &mut offset, &search_query));
+
+        if !context.worker_context_ref().is_busy() && !context.worker_context_ref().is_empty() {
+            context.worker_context_mut().start_next_job();
+        }
```
kamiyaa, line 33: This can be removed. I will move this into a different function later.

src/commands/help.rs
```diff
@@ -0,0 +1,90 @@
...
+    backend: &mut TuiBackend,
+    keymap_t: &AppKeyMapping,
+) -> JoshutoResult<()> {
+    context.flush_event();
+
+    let mut offset = 0;
+    let mut search_query = String::new();
+    let mut sort_by = 1;
+
+    while context.quit == QuitType::DoNot {
```
kamiyaa, line 23: You can change this to just `loop` and have a break condition (which you already have).
The user shouldn't be able to quit `joshuto` from here.

src/commands/help.rs
```diff
@@ -0,0 +1,90 @@
...
+                            KeyCommand::CursorMovePageUp => move_offset(&mut offset, -10),
+                            KeyCommand::CursorMovePageDown => move_offset(&mut offset, 10),
+                            KeyCommand::CloseTab | KeyCommand::Help => break,
+                            KeyCommand::ForceQuit => {
+                                eprintln!("quit");
+                                #[allow(unused_must_use)]
+                                {
+                                    quit::force_quit(context);
+                                }
+                            }
```
kamiyaa, line 65: Quit thing again. You can remove it,

src/commands/help.rs
```diff
@@ -0,0 +1,90 @@
...
+        if !context.worker_context_ref().is_busy() && !context.worker_context_ref().is_empty() {
+            context.worker_context_mut().start_next_job();
+        }
+
+        let event = match context.poll_event() {
+            Ok(event) => event,
+            Err(_) => return Ok(()),
+        };
+
+        if let AppEvent::Termion(event) = event {
```
kamiyaa, line 40: This should be a match statement so everything is processed.
ie. please use `process_noninteractive` for everything that is not user input.
This keeps background workers running.
example: https://github.com/kamiyaa/joshuto/blob/dev/src/util/input.rs#L49

kamiyaa:
Looks good overall, just some changes needed for the loop.
Should be able to be merged soon!

Great work!

PR #102, opened 2021-11-05
WIP: Add build/deploy workflow

### LINKED ISSUE #17: Publish binaries on new release using CI
So there's a cool project called [trust](https://github.com/japaric/trust) that you can setup in your project, so that every time you create a new release/create a new tag, it will automatically build binaries for you using travis ci and publish them to the releases tab. This is really nice for people who want to use rust projects but don't want to have to deal with its compile times :D

edit: Looks like that project might be abandoned, but the idea is still the same. Ripgrep and Alacritty both do some sort of similar thing for their projects, so we could always check those out too.

dufferzafar:
I was trying to fix #17 

The workflow was copied from https://github.com/sharkdp/fd/blob/master/.github/workflows/CICD.yml

But the last step (publishing a release) doesn't seem to be working.

The commit requires a tag, which I tried creating. But to no avail.

You can seem the actions log at my fork: https://github.com/dufferzafar/joshuto/runs/4119054327

### DIFF
--- a/.github/workflows/rust-linux-main.yml
+++ b/.github/workflows/rust-linux-main.yml
@@ -1,31 +1,138 @@
 name: Linux build
 
+env:
+  CICD_INTERMEDIATES_DIR: "_cicd-intermediates"
+
 on:
   push:
     branches: [main]
-  pull_request:
-    branches: [main]
+  workflow_dispatch:
 
 jobs:
-  check:
-    name: Rust Linux ${{ matrix.rust }}
-    runs-on: ubuntu-latest
+  build:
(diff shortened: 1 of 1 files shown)


kamiyaa:
Thanks for the PR! I'm not too familiar with Github's CI/CD so I will probably need todo some research before getting back to this PR if it's not ready xd

anni57779821988:
Could you explain what this line of code does? I'm not quite following.

scutuatua-crypto (approved):

scutuatua-crypto (approved):

scutuatua-crypto (approved):

PR #103, opened 2021-11-09
Path shortener

kennycallado:
Hi!

shortens the path when it is longer than area.width. I'm not sure if it's in a good place as I'm not an experienced programmer and I'm a newbie in rust.

### DIFF
--- a/src/ui/widgets/tui_topbar.rs
+++ b/src/ui/widgets/tui_topbar.rs
@@ -1,3 +1,4 @@
+use std::path::Component;
 use std::path::Path;
 
 use tui::buffer::Buffer;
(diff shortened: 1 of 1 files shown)


src/ui/widgets/tui_topbar.rs
```diff
@@ -31,8 +31,21 @@ impl<'a> Widget for TuiTopBar<'a> {
 
         if curr_path_str.len() > area.width as usize {
             if let Some(s) = self.path.file_name() {
+                /* path shortener */
+                let mut temp: Vec<&str> = curr_path_str.split("/").collect();
```
kamiyaa, line 35: Instead of using split here, I think we can use [Path.components](https://doc.rust-lang.org/std/path/struct.Path.html#method.components) method on `self.path` instead.
This will work on windows as well (not that joshuto supports windows currently), but it will also eliminate the problem of needing to check if the folder name is not empty string.

Something like:
```rust
let mut short_path = String::new();
for component in self.path.components() {
    match component {
        Component::RootDir => short_path.push('/')
        Component::Normal(s) => {
            // get first char and push it
            // push "/"
        }
    }
}
```

kamiyaa (commented):
Overall logic looks good though!

src/ui/widgets/tui_topbar.rs
inoshy: neat thinking!

src/ui/widgets/tui_topbar.rs
```diff
@@ -31,8 +32,22 @@ impl<'a> Widget for TuiTopBar<'a> {
...
+                    match component {
+                        Component::RootDir => short_path.push('/'),
+                        Component::Normal(s) => {
+                            let ch = s.to_string_lossy().chars().next().unwrap();
+                            short_path.push(ch);
+                            short_path.push('/');
+                        }
+                        Component::Prefix(_) => {}
+                        Component::CurDir => {}
+                        Component::ParentDir => {}
```
kamiyaa, line 46: Can be reduced to just
```rust
_ => {}
```

kamiyaa:
Thank you!

Issue #96, opened 2021-10-11
Add a "--choosefiles" like flag


is0n:
Ranger has this option where if `--choosefiles=/some/file/path` is added to `ranger`, the selected file's path will be written to the specified path instead of being opened it. This lets certain programs like vim/neovim to use ranger as a file manager.

PR #105, opened 2021-11-16
Add choosefiles feature

foxfriends:
Not sure if this is the best way to implement it, but I've been sitting on various iterations of this hack for a while now. I think it would close https://github.com/kamiyaa/joshuto/issues/96.

### DIFF
--- a/src/context/app_context.rs
+++ b/src/context/app_context.rs
@@ -19,6 +19,8 @@ pub struct AppContext {
     pub quit: QuitType,
     // event loop querying
     pub events: Events,
+    // choose file instead of opening it
+    pub choosefiles: bool,
     // app config
     config: config::AppConfig,
     // context related to tabs
@@ -38,7 +40,7 @@ pub struct AppContext {
 }
 
 impl AppContext {
-    pub fn new(config: config::AppConfig) -> Self {
+    pub fn new(config: config::AppConfig, choosefiles: bool) -> Self {
         let events = Events::new();
         let event_tx = events.event_tx.clone();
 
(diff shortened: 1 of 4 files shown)


src/context/app_context.rs
```diff
@@ -19,6 +19,8 @@ pub struct AppContext {
     pub quit: QuitType,
     // event loop querying
     pub events: Events,
+    // choose file instead of opening it
+    pub choosefiles: bool,
```
kamiyaa, line 23: Instead of having a `choosefiles` boolean here, I think we add a new variant to `QuitType` called `ChooseFiles` or something. That was kind of the idea of what `QuitType` is for; tell Joshuto to do things on quit.

Here, we can also add `Args` as a field of `AppContext`. So later on in `open_file.rs` we can check
`if context.args.choosefiles.is_some()`

src/main.rs
```diff
@@ -122,6 +124,14 @@ fn run_joshuto(args: Args) -> Result<(), JoshutoError> {
             }
         }
         QuitType::Force => {}
+        QuitType::Normal => {
+            if let Some(path) = args.choosefiles {
```
kamiyaa, line 128: Should be moved to `QuitType::ChooseFiles`.
Also you can move this into a separate function that returns `io::Result<()>`.
This way you don't need to use `unwrap()` and can just use `?` and handle the error here.

src/run.rs
```diff
@@ -91,5 +92,10 @@ pub fn run(
         }
     }
 
-    Ok(())
+    let curr_tab = context.tab_context_ref().curr_tab_ref();
```
kamiyaa, line 95: We can move this to `main.rs` and keep this function the same

kamiyaa (commented):
Overall logic seems fine. Just need some design changes. :)
Good work!

foxfriends:
Thanks for the feedback! I have made those changes. Let me know if anything else could be improved :)

kamiyaa:
LGTM, thanks!

Issue #67, opened 2021-06-07
File Preview Support


kamiyaa:
## This issue will track the discussion and progress of getting file previews to work

Just some food for thought on how to design this:
 - configuration format
   - how do we configure it so users can customize it to their liking
 - overhead costs
   - querying mimetype beforehand (during dirlist loading) or querying them on the fly (possibly increasing cursor movement delays)
 - passing files
   - how should we pass the file to preview scripts
 - showing previews
    - [tui-rs](https://github.com/fdehau/tui-rs) currently does not support ansi color code parsing so parsing colors might need to be implemented in-house
    - tui-rs also uses a different method of outputting text to the terminal. A buffer is used instead of directly interacting with the terminal screen. Any terminal output not tracked by the buffer is undetected and may create residual artifacts
    - will need to figure out how to isolate preview to the given window

DLFW:
And just some resulting thoughts:

tl;dr: Joshuto could start with a stripped down ranger approach.

* configuration format / passing files 
  Ranger just uses one bash file, giving it the full path and the geometry as arguments and taking the preview content from it's stdout and return code. Ranger's config description:
  ```bash
  # Meanings of exit codes:
  # code | meaning    | action of ranger
  # -----+------------+-------------------------------------------
  # 0    | success    | success. display stdout as preview
  # 1    | no preview | failure. display no preview at all
  # 2    | plain text | display the plain content of the file
  # 3    | fix width  | success. Don't reload when width changes
  # 4    | fix height | success. Don't reload when height changes
  # 5    | fix both   | success. Don't ever reload
  # 6    | image      | success. display the image $cached points to as an image preview
  # 7    | image      | success. display the file directly as an image

  # Meaningful aliases for arguments:
  path="$1"            # Full path of the selected file
  width="$2"           # Width of the preview pane (number of fitting characters)
  height="$3"          # Height of the preview pane (number of fitting characters)
  cached="$4"          # Path that should be used to cache image previews
  preview_images="$5"  # "True" if image previews are enabled, "False" otherwise.

  ```
  I like the simple and flexible approach. It would even be possible to let the user configure the script's/program's name and let him choose his preferred language. I think this sub-process call must anyway happen in a separate thread, so the performance shouldn't be problematic. 
  At least in a simple first approach, the exit codes 2 – 7 could be left out (the user could just “cat” the file to stdout), as well as the two last call arguments (which are for image previews).

[...]

PR #106, opened 2021-11-18
Add ANSI color codes support in file previews

krivahtoo:
Add ANSI color codes support in file preview using [ansi-to-tui](https://github.com/uttarayan21/ansi-to-tui) as suggested [here](https://github.com/fdehau/tui-rs/issues/304#issuecomment-827124643)

### Screenshot
Screenshot_20211118_155938

### Example

In `preview.sh`
```bash
# file extension "json"
jq --color-output . "${path}" && exit 0
# mime type "text/*"
bat --color=always --style="plain" -- "${path}" && exit 0
```

using `highlight` doesn't seem to work :(

related #67


kamiyaa:
LGTM! Thanks for the great work!

PR #109, opened 2021-12-02
add line numbers

### LINKED ISSUE #41: Feature request: line numbers
Hi there, any possibility to have line numbers, absolute or relative, similar to vim? So that we can do vim movement like <#-j> <#-k> <#-gg> etc.?

Thanks for this program!

Vinegret43:
closes #41

Adds numbers for every entry in the main window. Supports both absolute numbering (As shown on the first screenshot in the linked issue), and relative numbering (As shown on the second screenshot, again, see the issue)

Work in progress:
  - Refactoring and improving the code
  - Adding documentation

### DIFF
--- a/config/joshuto.toml
+++ b/config/joshuto.toml
@@ -12,6 +12,8 @@ show_hidden = false
 show_icons = true
 show_preview = true
 tilde_in_titlebar = true
+# no, absolute, relative
+line_nums = "no"
 
 [display.sort]
 # lexical, mtime, natural
--- a/config/joshuto.toml
+++ b/config/joshuto.toml
@@ -12,8 +12,8 @@ show_hidden = false
 show_icons = true
 show_preview = true
 tilde_in_titlebar = true
-# no, absolute, relative
-line_nums = "no"
+# none, absolute, relative
+line_number_style = "none"
 
 [display.sort]
 # lexical, mtime, natural
--- a/src/config/option/display_option.rs
+++ b/src/config/option/display_option.rs
@@ -19,10 +19,17 @@ pub struct DisplayOption {
     pub _show_preview: bool,
     pub _sort_options: SortOption,
     pub _tilde_in_titlebar: bool,
-    pub _line_nums: u8,
+    pub _line_nums: LineNumberStyle,
     pub column_ratio: (usize, usize, usize),
     pub default_layout: [Constraint; 3],
     pub no_preview_layout: [Constraint; 3],
+}
+
+#[derive(Clone, Copy, Debug)]
+pub enum LineNumberStyle {
+    None,
+    Relative,
+    Absolute,
 }
 
 impl DisplayOption {
@@ -70,12 +77,12 @@ impl DisplayOption {
         self._tilde_in_titlebar
     }
 
-    pub fn line_nums(&self) -> u8 {
+    pub fn line_nums(&self) -> LineNumberStyle {
         self._line_nums
     }
 
-    pub fn set_line_nums(&mut self, policy: u8) {
-        self._line_nums = policy;
+    pub fn set_line_nums(&mut self, style: LineNumberStyle) {
+        self._line_nums = style;
     }
 
     pub fn filter_func(&self) -> fn(&Result<fs::DirEntry, std::io::Error>) -> bool {
@@ -114,7 +121,7 @@ impl std::default::Default for DisplayOption {
             _show_preview: true,
             _sort_options: SortOption::default(),
             _tilde_in_titlebar: true,
-            _line_nums: 0,
+            _line_nums: LineNumberStyle::None,
             default_layout,
             no_preview_layout,
         }
--- a/src/config/option/display_option.rs
+++ b/src/config/option/display_option.rs
@@ -19,6 +19,7 @@ pub struct DisplayOption {
     pub _show_preview: bool,
     pub _sort_options: SortOption,
     pub _tilde_in_titlebar: bool,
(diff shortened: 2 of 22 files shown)


config/joshuto.toml
```diff
@@ -12,6 +12,8 @@ show_hidden = false
 show_icons = true
 show_preview = true
 tilde_in_titlebar = true
+# no, absolute, relative
+line_nums = "no"
```
kamiyaa, line 16: I think having something like `show_line_numbers` or something would be better. No need to shorten it as its just a config file that we probably won't be modifying often.

src/config/option/display_option.rs
```diff
@@ -19,6 +19,7 @@ pub struct DisplayOption {
     pub _show_preview: bool,
     pub _sort_options: SortOption,
     pub _tilde_in_titlebar: bool,
+    pub _line_nums: u8,
```
kamiyaa, line 22: You can make a separate enum for this instead of using arbitrary int values.
Something like:
```rust
pub enum LineNumberStyle {
    None,
    Relative,
    Absolute,
}
```
And the `no` option can be `none`

kamiyaa:
Overall, the changes looks like its on the right track! :+1: 
Not sure if you want this PR to only contain the line numbers and make another PR in the future on actually getting j, k, g, etc. to work.
Or if you're planning to get it all done in 1 PR

Vinegret43:
> Overall, the changes looks like its on the right track! +1 Not sure if you want this PR to only contain the line numbers and make another PR in the future on actually getting j, k, g, etc. to work. Or if you're planning to get it all done in 1 PR

Oh, i wanted to include moving by line numbers right into this PR to fix issue #41 entirely. Sorry if i made it a bit implicit

kamiyaa:
In terms of moving the cursor around relatively, I think you can just map to
`:cursor_move_up ` or `:cursor_move_down `.

This will prompt the user to put in a number and it will move up/down by that much.

For absolute movements, you will probably need a new command like `cursor_set_index` or something.

Let me know what you think :+1:

Vinegret43:
> This will prompt the user to put in a number and it will move up/down by that much

Don't see any sense in prompting the user, because in this case you will have to press `Enter`. In Vim, for example, you firstly put in a number and then enter just any command (Like `j`, `k`, `gg`). While you are entering the number, Vim displays it in the bottom panel. In my opinion, this is a better way, because the whole point of this is to speed up navigation, just one extra keypress may ruin all of that.

P.s. i already have a working implementation which works the same way i described, but i'll need to rewrite it because it's a mess. I also had to modify main loop in **run.rs** a bit so that when user enters a number it will call a special function for processing numbered commands. Is that okay? Sure, this way is a bit more complicated in terms of code, but at lest you don't have to prompt the user

Vinegret43:
Ok, so i made it. Modified **run.rs** a bit, added a new trait for `Command`, which is `NumberedExecute`. It allows you to make basically any command work in a different behavior when it is prefixed by a number. Tell me if you think these changes are ok and i should continue working on them or if i should implement it another way.

src/ui/widgets/tui_dirlist_detailed.rs
```diff
@@ -103,10 +134,19 @@ fn print_entry(
         drawing_width,
     );
 
-    let right_width = right_label.width();
-    buf.set_stringn(x, y, left_label, drawing_width, style);
+    // Drawing index (If not empty)
```
kamiyaa, line 137: I think theres some weird bugs going on here with calculating margins, spacing, etc.
But I'm guessing you're still ironing it out :+1:

kamiyaa (commented):
This is an interesting approach :thought_balloon: 
Honestly, I didn't have a design in mind for this feature without making lots of changes. So these changes being unobtrusive and more of an add-on to the existing logic, is really good. :heart: 

Some visual bugs with the printing that needs to be addressed. The printing code is pretty fragile so its expected.

There will probably be some worker-related bugs with the event loop inside `numbered_command` but that can be fixed without changing it.
I already have an idea in mind of how to change the event loop so all event loops can benefit from it (ie. moving most of the non-user input into `process_noninteractive`)

I think we can move forward with it :smile:

Vinegret43:
> Some visual bugs with the printing that needs to be addressed

Can you please describe these bugs? This may sound stupid, but i haven't encountered any visual bugs

kamiyaa:
Yeah sure :+1: 

It looks the right side margin has changed along with the left side changes.
Somehow they are in sync. Now its causing long extensions to show up when its not supposed (should be covered by the folder size).
I'm not too familiar with that part of the code because I didn't write it :eyes: (It was merged [here](https://github.com/kamiyaa/joshuto/pull/72))

# Actual
20211204-211415_swappy

# Original
20211204-211427_swappy

kamiyaa:
You can add some additional tests to test the function with indexes enabled with relative/absolute

Vinegret43:
Should be ready now if you think that everything is ok. Wanted to add these for tests for `print_entry` function but figured out that to make them you will need to construct `DisplayOption` and a lot of other structs from scratch, making these tests way too complicated

kamiyaa (approved):

kamiyaa:
LGTM! Thanks!

kamiyaa:
Looks like theres still some issues with the printing, but I will fix it.
Are you okay with the numbers being colored the same color as the file names?

image

kamiyaa:
Never mind, I found a way to fix it without having to use the same color

PR #110, opened 2021-12-12
Fix: DirList::need_update considers mtime change

DLFW:
1st commit: Fix: `JoshutoDirList::need_update` does not only consider explicit deprecation but also a changed mtime from the file system.

2nd commit: update also the left (parent dir) pane when switching tab

3rd commit: soft-reload also when opening a directory


kamiyaa:
Thanks!

PR #111, opened 2021-12-14
add dynamic number of columns

### LINKED ISSUE #104: [Feature Request] Dynamic Columns (Pans) configured from the joshuto.toml file
Ranger has the ability to change the number of pans, I took a look at the code and it seems that layout is hard coded am relatively new to rust, but I will try to add this in.

Vinegret43:
closes #104

Changes behavior of `column_ratio` property in **joshuto.toml**, now it can have length of 2 or 3. If it has length of 2 or first value is 0, parent view won't be displayed

### DIFF
--- a/config/joshuto.toml
+++ b/config/joshuto.toml
@@ -4,7 +4,7 @@ use_trash = true
 [display]
 automatically_count_files = false
 collapse_preview = true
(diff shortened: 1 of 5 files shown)


kamiyaa:
LGTM! Thanks!

PR #112, opened 2021-12-14
Update view on filesystem changes

DLFW:
AppContext owns an INotifyWatcher, which watches the visible directories
(max. three). The list of directories to watch is updated on every
event. If a filesystem change is announced, a "soft-update" of the
current view is issued.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,6 +161,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "245097e9a4535ee1e3e3931fcfcd55a796a44c643e8596ff6566d68f09b87bbc"
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.1"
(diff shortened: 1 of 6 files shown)


kamiyaa:
Looks like theres some problems on the Mac build.

Maybe this feature can be a Linux-specific feature for now, or we figure out a way to fix the Mac build.

DLFW:
It builds now for Linux and MacOS. I cannot test it on a Mac though. :) However, I guess for any other platform `notify` provides an implementation for, joshuto will now fail to build. If such platforms shall be supported to, we need some conditional in the code to just return a `NullWatcher` (which is a stub doing nothing), or a specific type in the struct. But again, I cannot test it and I prefer a build error over a runtime error.

I fear I need to leave that to people who actually build and use `joshuto` on other platforms.

kamiyaa:
Yep sounds good. I expect others will make PRs depending on if it works for their platforms or not. :+1:

kamiyaa (approved):

PR #113, opened 2021-12-20
Add script hooks for image previews

DLFW:
This adds two configuration options, `preview_shown_hook_script`
and `preview_removed_hook_script` in the preview-section where the
user can configure external scripts which are called when a file preview
is shown or when the file preview widget disappears, respectively.

The `preview_shown_hook_script` gets the path of the file being previwed
as first argument and then the x and y coordinate and the width an
height of the preview area as second to fifth parameters.

To get the geometry information back from the UI rendering process, a
new struct `RenderResult` has been introduced which is given as mutable
reference to the rendering function.

The main-loop then checks for a changed preview and calls the hook
scripts.

Additionally, documentation has been added how these hooks can be used
to implement image previews.


src/ui/mod.rs
```diff
@@ -1,5 +1,29 @@
+use std::path;
+
 mod tui_backend;
 pub mod views;
 pub mod widgets;
 
 pub use tui_backend::*;
+
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+pub struct Rect {
```
kamiyaa, line 10: Is there a reason for making our own `Rect` instead of using https://docs.rs/tui/0.16.0/tui/layout/struct.Rect.html

kamiyaa:
Tried it out and it looks like its working as expected.
I'm not sure how I feel about passing a mutable reference into a widget. Ideally I wanted widgets to just show stuff and not modify anything.

Would it be possible to make a function that calculates the areas (passing in a &AppContext should be all we need) and a function that does the preview stuff?
That way the widget can also make use of these functions.

It might be a bit tricky because some of the logic in the calculations are intertwined with previews and other things.

src/ui/mod.rs
DLFW: Yeah, lol, changed it two times back and forth. I went with this to stay more independent from the backend lib, as this would have been the only TUI structure which is used in the UI-independent logic. I can change that, if you like.

kamiyaa: No, its fine lol. Was just curious :+1:

docs/image_previews.md
```diff
@@ -0,0 +1,231 @@
...
+preview image with Überzug now. Joshuto offers two hooks which will be used for that.
+
+In your `~/.config/joshuto/joshuto.toml`, configure a script for each of these hooks: 
+
+```toml
+[preview]
+...
+preview_shown_hook_script = "~/.config/joshuto/on_preview_shown"
+preview_removed_hook_script = "~/.config/joshuto/on_preview_removed"
+```
```
kamiyaa, line 97: what do you think about naming them
```bash
preview_on_show_script
preview_on_remove_script
```

DLFW:
Totally agree, I also was not so happy about this dirty result-by-reference side-effect stuff. Keeping the UI a side-effect free function of the app state is desirable.

Moving the layouting-logic out of the render functions will be some work. It might be a good idea to do that for the complete UI logic to keep the code design consistent and slick. Would be quite some refactoring work though.

I'll see if I can come with something after Christmas.

kamiyaa:
I can merge it now and work on moving the logic out of the UI.

How does that sound?

docs/image_previews.md
DLFW: Hm, not my native language. For my ears it sounds like the scripts would provide a preview when some "show" or "remove" event happens. Not like the scipt is called _as a result_ of Joshuto's preview being shown or removed.

Maybe
```
on_preview_show_script
on_preview_remove_script
```
?

DLFW:
Sounds great to me! ;)

kamiyaa:
Thanks!

Issue #65, opened 2021-06-05
Support ueberzug for image preview 


siduck:
ranger and some other file managers support ueberzug , ueberzug can be used for viewing images , gif and this will work on any terminal , its not terminal specific like w3m :)

https://github.com/seebye/ueberzug

ranger's image preview : 

![image](https://user-images.githubusercontent.com/59060246/120878387-b9295180-c5d9-11eb-97fa-f76e51b80240.png)

kamiyaa:
Seems like `ueberzug` utilizes x11 features to accomplish this.
I no longer directly use x11 and it is slowly being phased out, so I'm not too motivated on getting this feature implemented.

However, if anyone else would like to implement this without introducing dependencies on x11, feel free!

siduck76:
Oh :( , almost most cli file managers I used did support ueberzug like clifm, lf, vifm etc! ueberzug depends on X yes , But Its the only image viewing protocol or such that works for every terminal unlike w3m , sixel etc , You cant 

![image](https://user-images.githubusercontent.com/59060246/120909798-72e5f800-c696-11eb-8245-5d583f0532f9.png)

DLFW:
Hi!
My 2 cents:
I guess an image preview feature will be wanted / needed / demanded by many users and it makes total sense to get it implemented. It's just super useful. On the other hand, I understand that avoiding certain dependencies (and X likely belongs to them) is necessary to keep the project slick, clean and flexible. Unfortunately, showing images in a terminal is something very special and will be hard to realize without some kind of dependency.

But I also guess that most users from the joshuto target audience will be fans of the  “do one thing and do it well” idea and won't have a problem to integrate other programs with some configuration and scripting effort.

So, just a rough idea: What about some kind of script hook when selecting a file?
As soon as we have some textual file preview implemented, we could have a feature that can call a script (on_selection) with the preview field’s starting coordinate, total height, and the number of lines of the textual preview (all in terminal character cells). That would allow that external script to use Ueberzug or another solution like [icat from Kitty](https://sw.kovidgoyal.net/kitty/kittens/icat.html) to show an image. When the file selection changes, another hook-script (“on_deselection”) could be called in where the user could remove the image. Only when this 2nd script terminates, the first one (on_selection) is called again with the currently selected file. (I guess the latter must also be called when joshuto terminates and when tabs change. Also geometry changes must be considered somehow.)

[...]

PR #115, opened 2021-12-21
refactor image preview via ueberzug

kamiyaa:
As mentioned in https://github.com/kamiyaa/joshuto/pull/113

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -2,18 +2,20 @@ use crate::commands::numbered_command;
 use crate::config::AppKeyMapping;
 use crate::context::{AppContext, QuitType};
 use crate::event::AppEvent;
-use crate::key_command::{AppExecute, Command, CommandKeybind};
+use crate::key_command::{AppExecute, CommandKeybind};
 use crate::preview::preview_default;
 use crate::tab::JoshutoTab;
 use crate::ui;
+use crate::ui::views;
 use crate::ui::views::TuiView;
 use crate::ui::RenderResult;
 use crate::util::input;
 use crate::util::to_string::ToString;
-use std::path;
+
 use std::process;
 use std::thread;
 use termion::event::{Event, Key};
+use tui::layout::Rect;
 
 pub fn run(
     backend: &mut ui::TuiBackend,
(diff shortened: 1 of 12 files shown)


src/run.rs
```diff
@@ -29,47 +31,15 @@ pub fn run(
         // trigger a preview of child
         preview_default::load_preview(context, backend);
     }
-    let mut last_preview_file_path: Option<path::PathBuf> = None;
+
+    let mut render_result: Option<RenderResult> = None;
```
kamiyaa, line 35: Instead of a struct with 2 `Option`, I changed it to an `Option` of the struct

src/run.rs
```diff
@@ -130,3 +100,56 @@ pub fn run(
     }
     Ok(())
 }
+
+fn process_preview_on_change(
```
kamiyaa, line 104: Moved most of the main loop code here

src/run.rs
```diff
@@ -130,3 +100,62 @@ pub fn run(
...
+    let new_render_result = views::calculate_preview(&context, layout[2]);
+
+    match new_render_result.as_ref() {
+        Some(new) => {
+            let should_preview = if let Some(old) = old_render_result {
+                new.file_preview_path != old.file_preview_path
+            } else {
+                true
+            };
+            if should_preview {
```
kamiyaa, line 133: preview if and only if different from previout preview (if it exists)

src/ui/views/tui_folder_view.rs
```diff
@@ -45,25 +41,11 @@ impl<'a> Widget for TuiFolderView<'a> {
...
-                        },
-                    },
-                }
-            };
+        let constraints = get_constraints(self.context);
+        let is_default_layout = constraints == &display_options.default_layout;
 
-        let layout_rect = if config.display_options_ref().show_borders() {
+        let layout_rect = if display_options.show_borders() {
+            let layout = calculate_layout_with_borders(area, constraints);
```
kamiyaa, line 48: We do rundandant calculations because we kind of have to render while figuring out the eventual layout

src/ui/views/tui_folder_view.rs
```diff
@@ -168,31 +138,24 @@ impl<'a> Widget for TuiFolderView<'a> {
...
-                        file_preview_path = Some(entry.file_path_buf());
-                        preview_area = Some(ui::Rect {
-                            x: rect.x,
-                            y: rect.y,
-                            width: rect.width,
-                            height: rect.height,
-                        })
-                    }
+        } else if curr_entry.is_some() {
+            let render_result = calculate_preview(self.context, layout_rect[2]);
```
kamiyaa, line 144: make use of helper to reduce code dupe

src/ui/views/tui_folder_view.rs
```diff
@@ -252,3 +215,106 @@ impl Intersections {
...
+    let preview_context = context.preview_context_ref();
+    let curr_tab = context.tab_context_ref().curr_tab_ref();
+
+    let child_list = curr_tab.child_list_ref();
+
+    let curr_list = curr_tab.curr_list_ref();
+    let curr_entry = curr_list.and_then(|c| c.curr_entry_ref());
+
+    if child_list.as_ref().is_some() {
+        None
```
kamiyaa, line 298: if we have a folder preview, preview for file should be `None`

kamiyaa:
@DLFW please take a look whenever you're free :) :+1:

DLFW:
Hey @kamiyaa, won't be able to have a proper look for the next week, but will love to have a look afterwards. ;)

kamiyaa:
Haha no problem! Happy holidays!

src/run.rs
```diff
@@ -37,9 +38,23 @@ pub fn run(
     while context.quit == QuitType::DoNot {
         backend.render(TuiView::new(context));
 
-        // compute preview render sizes
-        if let Ok(area) = backend.terminal_ref().size() {
-            render_result = process_preview_on_change(&context, area, render_result);
+        {
+            let config = context.config_ref();
+            let preview_options = config.preview_options_ref();
+            if let Some(shown_hook) = preview_options.preview_shown_hook_script.as_ref() {
```
DLFW, line 44: This handling has one downside: The shown-hood is only invoked if also the remove-hook is configured (and vice versa). While this is completely fine for the Überzug use case, this might be unexpected for users who have a different idea of what intersting stuff could be done with these hooks. If some user just wants to pop up a desktop notification with an image or whatever else, he won't need the remove-hook, but still needs to define one, just to get the shown hook-working. This might be a stupid example, but people might come up with more creative ideas of what can be done with it.

DLFW:
Thanks! :)
The changes look great!

Just a few remarks:

1) As mentioned above, the hooks should be independent of each other to allow other use cases and to avoid breaking user expectations, IMHO.

2) With this new design, the name `RenderResult` (and `render_result`) doesn't fit anymore. It's not the result of the rendering process as a whole anymore, but only specifically of the preview-area-calculation. Maybe it should be renamed to just  `Preview`. It's the result of `calculate_preview` after all, so that name would and should make sense.

3) `calculate_layout(_with_borders)` is now called twice per cycle. I have no idea how to avoid that except refactoring the UI calculation in a way that we calculate the layout before the rendering, keep this UI state in a separate structure, and pass it to both, the rendering process and the hook logic. But that would be quite some work, of course. Good enough for now, I guess.

src/run.rs
kamiyaa: Yeah, I can split it up so they aren't dependent on each other :+1:

kamiyaa:
> With this new design, the name RenderResult (and render_result) doesn't fit anymore. It's not the result of the rendering process as a whole anymore, but only specifically of the preview-area-calculation. Maybe it should be renamed to just Preview. It's the result of calculate_preview after all, so that name would and should make sense.

Yeah, I can rename it to something like `PreviewArea`

DLFW:
Thanks!
Guess #65 and #67 can both be closed.

PR #118, opened 2021-12-30
Continuous scrolling 

DLFW:
The scrolling behavior is changed from “paging” to a continuous
scrolling. Joshuto keeps a buffer from the cursor to each end of the
list, which is configured by `[display] scroll_offset`. If the terminal
height is too small to keep the distance, the buffer is set to a value
that assures that the cursor is at least as close to the end the user is
scrolling towards as to the other end of the visible list.

If the window is resized and the cursor jumps out of scope, the viewport
is adjusted when changing the index next time.

Possible improvements:
* Issue a viewport update on terminal geometry change
* When scrolling down to the bottom, don't allow an empty section
  beneath the last entry

### DIFF
--- a/src/context/app_context.rs
+++ b/src/context/app_context.rs
@@ -1,5 +1,6 @@
 use std::collections::HashSet;
 use std::sync::mpsc;
+use tui::layout::Rect;
 
 use crate::config;
 use crate::context::{
@@ -18,6 +19,11 @@ pub enum QuitType {
     Force,
     ToCurrentDirectory,
     ChooseFiles,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct UiContext {
+    pub layout: Vec<Rect>,
 }
 
 pub struct AppContext {
@@ -42,6 +48,8 @@ pub struct AppContext {
     preview_context: PreviewContext,
     // context related to command line
     commandline_context: CommandLineContext,
+    // user interface context; data which is input to both, the UI rendering and the app state
+    ui_context: UiContext,
     // filesystem watcher to inform about changes in shown directories
     #[cfg(target_os = "linux")]
     watcher: notify::INotifyWatcher,
@@ -81,6 +89,7 @@ impl AppContext {
             message_queue: MessageQueue::new(),
             worker_context: WorkerContext::new(event_tx),
             preview_context: PreviewContext::new(),
+            ui_context: UiContext { layout: vec![] },
             commandline_context,
             config,
             watcher,
@@ -170,6 +179,13 @@ impl AppContext {
     }
     pub fn preview_context_mut(&mut self) -> &mut PreviewContext {
         &mut self.preview_context
+    }
+
+    pub fn ui_context_ref(&self) -> &UiContext {
+        &self.ui_context
+    }
+    pub fn ui_context_mut(&mut self) -> &mut UiContext {
+        &mut self.ui_context
     }
 
     pub fn worker_context_ref(&self) -> &WorkerContext {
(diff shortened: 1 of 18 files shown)


src/run.rs
```diff
@@ -36,27 +39,25 @@ pub fn run(
...
-                    preview_area,
-                    preview_options.preview_shown_hook_script.as_ref(),
-                    preview_options.preview_removed_hook_script.as_ref(),
-                );
-            }
+            // render the ui
+            backend.render(TuiView::new(context));
+
+            // invoke preview hooks, if appropriate
+            preview_area = process_preview_on_change(&context, preview_area);
```
DLFW, line 51: The layout calculated in l45 before is now reused in here.

src/context/app_context.rs
```diff
@@ -20,6 +21,11 @@ pub enum QuitType {
     ChooseFiles,
 }
 
+#[derive(Clone, Debug, PartialEq)]
```
DLFW, line 24: Maybe this should be moved in it's own module at some time. But it's so small,  I thought that could wait....

src/run.rs
```diff
@@ -36,27 +39,25 @@ pub fn run(
     let mut preview_area: Option<PreviewArea> = None;
 
     while context.quit == QuitType::DoNot {
-        backend.render(TuiView::new(context));
+        // do the ui
+        if let Ok(area) = backend.terminal_ref().size() {
+            // pre-calculate some ui attributes
+            calculate_ui_context(context, area);
```
DLFW, line 45: I introduced this `UiContext` struct which can be used in general to store all that UI attributes which are calculate before the render process to have it accessible also for the app logic.

DLFW:
Just realized there is a misbehavior when re-entering a dir with having the cursor at an index > viewport-height: Joshuto uses a viewport-index of 0 in that case and the cursor is out of scope. I will have a look the next days.

DLFW:
Was a small thing. I observed the problem in a virtual file system and it turns out that Joshuto used `create_dirlist_with_history` there. All situations a dirlist was replaced by `create_dirlist_with_history` had the same problem. Should be fixed sufficiently now. There might be edge cases where a long dir list gains or looses a lot entries due to file system changes where the cursor ends up outside of the viewport. But I think it's good to merge now.

src/fs/dirlist.rs
```diff
@@ -10,25 +12,31 @@ pub struct JoshutoDirList {
...
     pub contents: Vec<JoshutoDirEntry>,
     pub metadata: JoshutoMetadata,
-    pub index: Option<usize>,
-    pub viewport_index: usize,
+    /// The cursor position in this dir list
+    index: Option<usize>,
+    /// The index in this dir list to start with when rendering the list
+    viewport_index: usize,
     _need_update: bool,
+    scroll_offset: usize,
```
kamiyaa, line 20: What do you think about putting `scroll_offset` somewhere else?
Ie. `UiContext`, or just leaving it in `DisplayOptions`
This will reduce the amount of memory used because we don't really need multiple copies of `scroll_offset` everywhere.

kamiyaa:
Sorry for the delay. Overall the changes looks good! I tried it out and it looks to be working fine :muscle:

src/fs/dirlist.rs
DLFW: Should be possible.
* `UiContext` does not feel like the right place for me. I thought about it as a place to store pre-calc UI information from the rendering process. On the other hand, that would be the cheapest solution.
* `DisplayOptions` is the place where it belongs, but that would mean that we need to pass DisplayOptions (or a clone) each time we call `set_index`.

I will think about it. ;)

DLFW:
No prob! Happy new year! :)

src/fs/dirlist.rs
DLFW: Done. Passing `DisplayOptions` (or a clone) to `set_index`.

kamiyaa (approved):

kamiyaa:
LGTM! Thanks!

PR #120, opened 2022-01-06
implement 'cd -', refactor directory changing code

### LINKED ISSUE #81: implement cd -
should behave similar to ranger's cd -

Vinegret43:
Closes #81

Added ability to change to last working directory by typing `cd -`. If you haven't changed any directories yet, it will not cd

Refactored directory changing code because it was a bit confusing to me when i was working with it. The changes include:
  - Merge `commands/parent_directory.rs` and `commands/change_directory.rs` into one file, `change_directory.rs`, which also contains the new code for `cd -`. Not very sure about this change, maybe it's a better idea to split them into 3 different files. Let me know your thoughts
  - Functions from these files (`change_directory` and `parent_directory`) were using some "helper" functions. Looks like it's just legacy code which was left when joshuto was using structs for commands instead of functions. This was really confusing to me, so i just moved code from these helper functions right into the main ones, hope this is ok


kamiyaa:
> - Functions from these files (change_directory and parent_directory) were using some "helper" functions. Looks like it's just legacy code which was left when joshuto was using structs for commands instead of functions. This was really confusing to me, so i just moved code from these helper functions right into the main ones, hope this is ok

Yep, that is correct. I'm totally fine with the changes.

> Merge commands/parent_directory.rs and commands/change_directory.rs into one file, change_directory.rs, which also contains the new code for cd -. Not very sure about this change, maybe it's a better idea to split them into 3 different files. Let me know your thoughts

I'm fine with either.

Overall looks good!

PR #121, opened 2022-01-09
Fix: invoke correct hook script

DLFW:
When the preview disappears, the "shown" hook script was spawned instead
of the "removed" hook script.

This mistake was not obvious with the preview-image recipe from the docs
which works also with the shown-script on remove.


kamiyaa:
Oops, thanks for the catch!

PR #128, opened 2022-01-19
fix: disarm spawn bomb

DLFW:
The preview-removed-hook was called in each main loop cycle if no
preview area was shown and a removed-hook was configured.


kamiyaa (approved):

kamiyaa:
Good work, thanks!

DLFW:
Thanks for the quick merge! :+1:

PR #131, opened 2022-01-23
fix doc of shown-hook for icat

DLFW:
When changing from an image file to another file type, the last image
was not cleared with the given description.


kamiyaa:
Thanks!

PR #132, opened 2022-01-25
Fix: remove image preview on bulk rename

krivahtoo:


kamiyaa:
Looks good overall, if you could just run cargo fmt to fix the build.
I'll merge it after it passes :+1:

krivahtoo:
Done

PR #135, opened 2022-01-27
code cleanup

DLFW:
Remove dead code and unused imports around preview features.

### DIFF
--- a/config/joshuto.toml
+++ b/config/joshuto.toml
@@ -25,7 +25,6 @@ reverse = false
 
 [preview]
 max_preview_size = 2097152 # 2MB
-preview_images = false
 preview_script = "~/.config/joshuto/preview_file.sh"
 
 [tab]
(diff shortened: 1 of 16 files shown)


DLFW:
Two configuration options seem to be unused:
* `[Display] show_preview`
* `[Preview] preview_images`

Can I remove these?

The latter is only given to the `preview` script, but as this script is not really resonsponsible for dealing with image previews, this should not be needed. Also, just handing over a configuration option to an external script seems pointless.

kamiyaa:
Yeah we can remove those options

DLFW:
Beside the two mentioned config options, I removed some unused functions and unused imports. Three build warnings are left: Two simple unsused functions where I thought there might be some need in the near future and one unchecked result of `event_tx.send` which I did not want to silently ignore without having a closer look.

Hope removing the unused code was fine.

kamiyaa (approved):

PR #139, opened 2022-02-23
Autocomplete first words as commands

sushi-shi:
Good day! Some commands are multi word, say, `sort reverse`,  so this commit won't be able to autocomplete those after a space is pressed. 
To solve this we can either append those command matches to a `FilenameCompleter`, or make all commands to be one word only.
Also I don't like how constants need to be specified twice now, it can be improved with a macro, but I wonder whether there is a more general solution.

### DIFF
--- a/src/key_command/constants.rs
+++ b/src/key_command/constants.rs
@@ -1,61 +1,85 @@
-pub const CMD_HELP: &str = "help";
+use rustyline::completion::Pair;
 
-pub const CMD_QUIT: &str = "quit";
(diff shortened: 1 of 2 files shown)


kamiyaa:
LGTM!

kamiyaa:
> Also I wonder whether it would make sense to fix all `clippy` warnings (there are ~50 of those) and add CI which will verify that new commits do not introduce new warnings as well.

Imo the clippy suggestions aren't necessarily better in some cases so I wouldn't worry about it.
Sometimes it suggests code that doesn't compile :eyes: but that might have been improved.

Issue #138, opened 2022-02-23
[Feature Request] Interactive search


sushi-shi:
Good day!

I would want to implement an interactive search, as I miss it so much after switching from Ranger.
Here is an [example](https://github.com/sushi-shi/test/blob/master/out.gif).

I tried to [implement](https://github.com/sushi-shi/joshuto/commit/c640cd22c21140eec5405cc89b5ab2d48b53936a) it by making an iterator (kind of), which returns `String` from the current `line_buffer`, parses it and runs `search_string` if the command matches. 
Sadly, after that I realized that this iterator mutably borrows `AppContext` and `search_string` also needs to mutably borrow it.
So it doesn't work... (Also it is a bit wasteful, as it needs to `clone` on each key press). I _can_ make this approach work by wrapping required fields into `Rc<RefCell<_>>` but it doesn't feel like a good solution. As another solution, I think, I can add a callback, which `read_and_execute` and `user_input` functions will accept and call whenever `line_buffer` changes, it feels like it should work, however, before I proceed I would like to ask for your opinion first.

Thank you for making such a nice tool!

sushi-shi:
Also it feels like a `:command [args]` string shouldn't be a command and can be parsed elsewhere. It would also allow to autocomplete commands (which is the next thing I want to tackle)

kamiyaa:
What do you mean by interactive search?

Also yeah, I'm fine with `:command` not being a command and being parsed elsewhere

sushi-shi:
Being able to search as you type!

kamiyaa:
I think interactive search can already be done via fzf or are you thinking of something else?

sushi-shi:
As I write a search pattern I want joshuto to interactively (i.e. without press Enter) select the first best match. I did write a small PR which implements it

kamiyaa:
> As I write a search pattern I want joshuto to interactively (i.e. without press Enter) select the first best match. I did write a small PR which implements it

Ah I see, that makes sense. Will this conflict with the line number code?
https://github.com/kamiyaa/joshuto/commit/5f7bd3889dfc054308c19f655aa951946cfb299c

Where you can start typing numbers to move the cursor.

kamiyaa:
Closing via https://github.com/kamiyaa/joshuto/pull/140

PR #140, opened 2022-02-23
[#138] Add incremental search

sushi-shi:
Implementing https://github.com/kamiyaa/joshuto/issues/138


src/error/mod.rs
```diff
@@ -4,4 +4,4 @@ mod error_type;
 pub use self::error_kind::JoshutoErrorKind;
 pub use self::error_type::JoshutoError;
 
-pub type JoshutoResult<T> = Result<T, JoshutoError>;
+pub type JoshutoResult<T = ()> = Result<T, JoshutoError>;
```
kamiyaa, line 7: What is this doing exactly? I'm not too familiar with this syntax lol

src/key_command/command.rs
```diff
@@ -52,6 +52,7 @@ pub enum Command {
 
     SearchGlob(String),
     SearchString(String),
+    SearchInc(String),
```
kamiyaa, line 55: I think we can call this `SearchIncrement`, `SearchIncremental`, or `IncrementalSearch`.
More descriptive and the user won't see it anyways.

src/key_command/impl_appexecute.rs
```diff
@@ -70,7 +70,13 @@ impl AppExecute for Command {
...
             Self::TouchFile(arg) => touch_file::touch_file(context, arg.as_str()),
             Self::SearchGlob(pattern) => search_glob::search_glob(context, pattern.as_str()),
-            Self::SearchString(pattern) => search_string::search_string(context, pattern.as_str()),
+            Self::SearchString(pattern) => Ok(search_string::search_string(
+                context,
+                pattern.as_str(),
+                false,
+            )),
+            // incremental_execute already has taken care of it
+            Self::SearchInc(_) => Ok(()),
```
kamiyaa, line 79: A little confused why this has to return `Ok(())`

sushi-shi: We call `incremental_execute` on each key press, so even before `Enter` is pressed the cursor will be one the selected word. And as `incremental_execute` always starts from index 0, this operation will be a no-op

src/key_command/command.rs
sushi-shi: Done!

kamiyaa:
LGTM! Is this ready for merge?

PR #142, opened 2022-02-23
Console shortcuts

sushi-shi:
Add shortcuts such as Ctrl+W, Ctrl+D, etc. to the textfield.
Also make auto completion work a bit better by moving cursor and then trying to complete instead of splitting the word in two.


kamiyaa:
LGTM Thanks!

PR #143, opened 2022-02-23
Move half page

sushi-shi:
`ctrl+d` and `ctrl+u` are used in vim to move half page down and up accordingly. Add support for the same thing.


sushi-shi:
Argh.., it also contains commits from the other PR. I wonder whether this one can be accepted with the previous one being closed. Or will everything work out just fine? I'm not sure

kamiyaa:
Looks like it can be merged :+1:

PR #144, opened 2022-02-24
Fix clippy warnings

sushi-shi:
It might not be necessary, but I still fixed all clippy warnings. It might also be reasonable to add clippy stage into CI, so that no warnings are introduced further on

### DIFF
--- a/src/context/worker_context.rs
+++ b/src/context/worker_context.rs
@@ -29,7 +29,8 @@ impl WorkerContext {
     // worker related
     pub fn push(&mut self, thread: IoWorkerThread) {
         self.worker_queue.push_back(thread);
-        self.event_tx.send(AppEvent::IoWorkerCreate);
+        // error is ignored
+        let _ = self.event_tx.send(AppEvent::IoWorkerCreate);
     }
     pub fn is_busy(&self) -> bool {
         self.worker.is_some()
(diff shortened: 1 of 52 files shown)


sushi-shi:
This small script changed quite a few files. I wonder whether this is okay or it might be better to change them incrementally.
```bash
[sheep]$ set files (rg 'JoshutoResult<\(\)>' --files-with-matches)
[sheep]$ for file in $files; sd -s 'JoshutoResult<()>' 'JoshutoResult' $file ; end
```

src/context/worker_context.rs
```diff
@@ -29,7 +29,8 @@ impl WorkerContext {
     // worker related
     pub fn push(&mut self, thread: IoWorkerThread) {
         self.worker_queue.push_back(thread);
-        self.event_tx.send(AppEvent::IoWorkerCreate);
+        // error is ignored
+        let _ = self.event_tx.send(AppEvent::IoWorkerCreate);
```
sushi-shi, line 33: I also wonder whether here should be `unwrap`.

kamiyaa: Should be fine

kamiyaa (approved):

kamiyaa:
Thanks for the work!

PR #149, opened 2022-03-01
Add docker to build the binary and copy it to a clean busybox image

kennycallado:
Not sure if it's useful but the Dockerfile compiles joshuto in a rust container and copy the binary into a busybox image. It results in a slim image with joshuto. 

  

Instructions: 

  

``` bash 

docker build -t joshuto:0.9.2 . 

``` 

``` bash 

docker run --rm -it -u $(id -u):$(id -g) -v $(pwd):/root joshuto:0.9.2 

``` 

And inside the container: 

  

``` bash 

joshuto 

```

### DIFF
--- a/.Dockerignore
+++ b/.Dockerignore
@@ -1,0 +1,9 @@
+config
+docs
+target
+Cargo.lock
+Docker*
+LICENSE
+README.md
+screenshot.png
+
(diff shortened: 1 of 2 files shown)


kamiyaa:
Yeah, this will probably be useful in the future :+1:

Issue #147, opened 2022-03-01
Only the first option in a mimetoml is picked up


sushi-shi:
I only have `sxiv` installed, but this `mimetype.toml` fails to open files with `No such file or directory (os error 2)`

```
image_default	= [
	{ command = "qimgv", args = [ "--" ], fork = true, silent = true },
	{ command = "sxiv", args = [ "--" ], fork = true, silent = true },
]
```

If you flip the commands everything works fine.

Possibly because of this:
```bash
[home]$ uname -a
Linux nixos 5.10.88 #1-NixOS SMP Wed Dec 22 08:31:00 UTC 2021 x86_64 GNU/Linux
```

I'll have a look later, when I have more free time

kamiyaa:
Yeah so if you just open a file with enter or something, it will just run the first option.
If you want to open with another option, you can use `:open_with`

PR #150, opened 2022-03-01
Try all options when opening files

sushi-shi:
Resolves https://github.com/kamiyaa/joshuto/issues/147


sushi-shi:
I still am not happy with those long `if else` chains. 
This macro can be used to make them more readable, I think:
https://docs.rs/kiam/latest/kiam/macro.when.html

kamiyaa (approved):

kamiyaa:
Looks good, thanks!

PR #152, opened 2022-03-02
Fixed copy instruction to use the binary from the builder container

kennycallado:
Sorry! I forgot to double check 😕


kamiyaa:
No problem! Thanks :+1:

Issue #151, opened 2022-03-02
theme.toml does not parse hex colors


Beethoven-n:
`str_to_color()` [in style.rs](https://github.com/kamiyaa/joshuto/blob/18d4a4c8a7c5060ed0f5094c958afe4f800e91d4/src/config/theme/style.rs#L64) only has a method to parse Rgb values, when colors_transform has methods to convert hex colors into Rgb colors.

Before the Rgb parsing, it should be possible to convert a hex color into an Rgb with `rgb.from_hex_str`, then pass that into the Rgb parsing method. (there's probably a better way to do that)

kamiyaa:
Shouldn't be too hard to implement. But currently, I'm a bit busy and probably won't get around to working on this at the moment

Beethoven-n:
would've done a pull request about it if i was comfortable enough with rust to know that the way i thought it should be done was the correct way

kamiyaa:
Closed via https://github.com/kamiyaa/joshuto/pull/154

PR #154, opened 2022-03-03
Fix for issue #151

Beethoven-n:
Consulted with a friend who knows Rust better than me and was able
to implement a fix for #151 pretty quickly after that point.

also thanks for having 2 different Rgb libraries active at the same time lmao

### DIFF
--- a/src/config/theme/style.rs
+++ b/src/config/theme/style.rs
@@ -69,7 +69,7 @@ impl RawAppStyle {
                 let g = rgb.get_green() as u8;
                 let b = rgb.get_blue() as u8;
                 style::Color::Rgb(r, g, b)
-            }
+            },
             s if s.is_empty() => style::Color::Reset,
             s => match s.parse::<Rgb>() {
                 Ok(rgb) => {
--- a/src/config/theme/style.rs
+++ b/src/config/theme/style.rs
@@ -60,6 +60,13 @@ impl RawAppStyle {
             "light_cyan" => style::Color::LightCyan,
             "white" => style::Color::White,
             "reset" => style::Color::Reset,
+            s if s.starts_with('#') => {
+                let rgb = Rgb::from_hex_str(s).unwrap();
+                let r = rgb.get_red() as u8;
+                let g = rgb.get_green() as u8;
+                let b = rgb.get_blue() as u8;
+                style::Color::Rgb(r, g, b)
+            },
             s if s.is_empty() => style::Color::Reset,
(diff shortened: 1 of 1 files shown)


src/config/theme/style.rs
```diff
@@ -60,6 +60,13 @@ impl RawAppStyle {
             "light_cyan" => style::Color::LightCyan,
             "white" => style::Color::White,
             "reset" => style::Color::Reset,
+            s if s.starts_with('#') => {
+                let rgb = Rgb::from_hex_str(s).unwrap();
```
kamiyaa, line 64: Rather than unwrap, we should handle the error and return `style::Color::Reset`

kamiyaa (changes requested):

src/config/theme/style.rs
Beethoven-n: i tried that with a `match` statement like you did, but it turns out that when you do that you get errors because it gets a different result on error

Beethoven-n: i'm not exactly sure What error to parse, is all

kamiyaa: Try something like and see if that works
```rust
s if s.starts_with('#') => {
    let rgb = match Rgb::from_hex_str(s) {
        Ok(s) => s,
        _ => return style::Color::Reset,
    };
    let r = rgb.get_red() as u8;
    let g = rgb.get_green() as u8;
    let b = rgb.get_blue() as u8;
    style::Color::Rgb(r, g, b)
},
```

Beethoven-n: yeah that works. will push this

Beethoven-n: yeah that works. will push

kamiyaa (approved):

kamiyaa:
If you could run `cargo fmt` on the project, I'll merge it :+1: 

Thanks for the contribution

Beethoven-n:
reformatted and ready for merge!

Beethoven-n:
a missing comma was detected by the workflow, so i've pushed a new commit to fix it :facepalm:

kamiyaa:
> reformatted and ready for merge!
> edit: thanks for the guidance! this is my first actual contribution to any project, period. it's not even in a language i'm fluent in! i'm super happy that this worked out as well as it did.

Glad I was part of your first contribution :heart:

Beethoven-n:
your workflows will complain about a misplaced comma, sorry lmao i only realized *after* you merged that it needs to be fixed

PR #160, opened 2022-03-22
Attach binaries to new releases

### LINKED ISSUE #17: Publish binaries on new release using CI

kennycallado:
Hi,

Action to build the binaries. It's triggered on new releases or tags (only those with v before the version number) and attaches the binaries to them.

Fix #17.


.github/workflows/autometed-build.yml
```diff
@@ -0,0 +1,155 @@
+name: Automated build
```
kamiyaa, line 1: File should be `automated-build` right?

kennycallado: Totally right 🙃

kennycallado:
I should have given more details with PR, but I'm not very proud of my English 😅. 

The v prefix is not required if you release from github. You can create a new release with or without v and the action is going to start the build process to attach the binaries. 

There is another way to proceed. You can tag locally `git tag v0.9.3` (note the v) and push this tag to github `git push --tags`. Github stores the tag as expected, but the action is going to create a release (which is editable) and attach the binaries to it. 

However, there is the option to tag without creating a release. If the tag doesn't have v at the beginning, it is going to be stored but it won't fire the action. 

Then you can have tags like '0.9.3.1' which is not a release, it is just a tag, and others like 'v0.9.4' which triggers the action and creates a release. But you are also able to create a release like '0.9.3.1' (no v) if you release from github. Then to keep consistency, I think could be better to prefix the releases with v and tags without v.

Released from github:
screenshot-1

Released from tag:
screenshot-2

Just a tag:
screenshot-3

I hope I have explained it well.

kamiyaa:
I see! Thanks for the explanation!

PR #165, opened 2022-04-14
Use notify::RecommendedWatcher to simplify code and support BSD platforms

unrelentingtech:
Along with https://github.com/notify-rs/notify/pull/399 this fixes build on FreeBSD.


kamiyaa:
LGTM! Thanks!

PR #166, opened 2022-04-30
use cli-clipboard for path copy

ztlevi:
Hi, 

I have xclip installed on my MacOS and it hangs during path copy. Use the clipboard library for better cross platform support.

### DIFF
--- a/src/commands/file_ops.rs
+++ b/src/commands/file_ops.rs
@@ -121,6 +121,10 @@ fn copy_string_to_buffer(string: String) -> JoshutoResult {
             format!("printf '%s' '{}' | {} -ib 2> /dev/null", string, "xsel"),
         ),
         (
+            "pbcopy",
+            format!("printf '%s' '{}' | {} 2> /dev/null", string, "pbcopy"),
+        ),
+        (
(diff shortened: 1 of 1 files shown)


ztlevi:
Tested on my mac but haven't tested on Linux since the dockerfile doesn't seem to work.

kamiyaa:
I would like to avoid bringing in dependencies like cli-clipboard because it introduces dependencies for x11 and wayland libraries on Linux. Ie. Joshuto will fail to compile if either is missing.

Instead, I would like to keep the approach of using cli commands to copy things to clipboard.

If the existing code does not work for Mac, you can add a Mac specific command when compiling for Mac.

Let me know if you have any questions!

ztlevi:
Got it. I don't feel like cli-clipboard needs additionaly linux library while [rust-clipboard](https://github.com/aweinstock314/rust-clipboard) did mention xorg-dev needs to be installed.

But based on your concern, I added `pbcopy` command to support MacOS clipboard. pbcopy comes with mac by default

kamiyaa:
LGTM! Thanks!

PR #171, opened 2022-05-15
add ctrl+h as hidden files toggle

Diaoul:
Pretty much like everywhere else on linux

Awesome project, I just discovered it :heart:


kamiyaa:
LGTM thanks!

PR #173, opened 2022-05-22
Add GitHub workflow for Clippy

dmarcoux:


dmarcoux:
Hey, I am not sure if you prefer to have an extra workflow for Clippy. I believe it's a better solution than adding Clippy to the existing workflows since this would be duplicating the reports generated by Clippy.

kamiyaa:
Yeah I think having them as separate files is good. 👌

PR #177, opened 2022-06-18
Add `flat` command

### LINKED ISSUE #161: `flat` command
Ranger has a command `flat` which allows a heavily nested folder to be browsed as a flattened list.
I think this is a nice feature to add to Joshuto; ideally it would be possible to bulk-rename files in this format too.

## Example

```
A/
    B.txt
    C/
        D.txt
        E.txt
```

Is currently displayed as:

```
A
```

With flat mode enabled, it would be displayed like this:

```
A/B.txt
A/C/D.txt
A/C/E.txt
```

krivahtoo:
closes #161

### DIFF
--- a/src/commands/flat.rs
+++ b/src/commands/flat.rs
@@ -12,7 +12,7 @@ fn _is_hidden(entry: &walkdir::DirEntry) -> bool {
     entry
         .file_name()
         .to_str()
-        .map(|s| s.starts_with("."))
+        .map(|s| s.starts_with('.'))
         .unwrap_or(false)
 }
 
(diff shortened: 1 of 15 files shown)


src/commands/flat.rs
```diff
@@ -0,0 +1,82 @@
...
+            if options.show_hidden() {
+                true
+            } else {
+                !_is_hidden(e)
+            }
+        })
+        .filter(|e| {
+            !e.as_ref()
+                .ok()
+                .unwrap()
```
kamiyaa, line 37: This crashes
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/commands/flat.rs:37:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I think the cause for this is when you don't have enough permissions for specific folders. ie. `/root`.

kamiyaa (changes requested):
Looks good overall, just a minor bug.

Thanks!

src/commands/flat.rs
krivahtoo: fixed

kamiyaa:
Thanks!

PR #183, opened 2022-07-13
File type filter for key maps

foxfriends:
With the recent change that implemented `quit --output-selected-files`, the `open` command no longer causes an exit the way `--choosefiles` did. I liked using it that way, where "opening" a file when using Joshuto as a file picker was doing the `quit --output-selected-files` action. 

To get that back, I have added a new field `filetype` to the keymap file, so I can write something like this:

```toml
[[default_view.keymap]]
command = "open"
keys = ["arrow_right"]
filetype = "directory"
[[default_view.keymap]]
command = "quit --output-selected-files"
keys = ["arrow_right"]
filetype = "file"
```

Now each key can do a different thing based on whether the entry is a directory or a regular file. This might even make it possible to remove that special case in `open` for directories, and change `arrow_right` on directory to do a `cd` instead, allowing `open` on a directory to do something even more special, if so desired (open in OS file explorer/open as project in VS code, etc)

### DIFF
--- a/src/commands/numbered_command.rs
+++ b/src/commands/numbered_command.rs
@@ -51,18 +51,38 @@ pub fn numbered_command(
                     Event::Key(Key::Char(c)) if c.is_numeric() => {
                         prefix.push(c);
                     }
-                    key => match keymap.default_view.get(&key) {
-                        Some(CommandKeybind::SimpleKeybind(command)) => {
-                            return command.numbered_execute(num_prefix, context, backend, keymap);
+                    key => {
+                        match keymap.default_view.get(&key) {
+                            Some(CommandKeybind::SimpleKeybind(filetypes)) => {
+                                let command = context
+                                    .tab_context_ref()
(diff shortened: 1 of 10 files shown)


kamiyaa:
Thanks for the PR!

Apologies for removing your feature! However, I'm not sure how I feel about this change.

I think instead, we can just add back the original functionality.
Having a `--file-chooser` option or something that would do the same thing (or something similar) as `quit --output-selected-files`

What do you think?

> if so desired (open in OS file explorer/open as project in VS code, etc)

And to address this, this is already possible via `open_with` on a directory (though it might be a bit more tedious)

foxfriends:
Yeah, if this is a better way to do it, that works for me too. I can continue using my fork for now, so no rush either way

kamiyaa:
Closed via 12e63afdd0111918fcef426a3b00718411e8e7c3

Please let me know if there are any discrepancies between the new option and old option

PR #191, opened 2022-08-06
Sort options individual per tab

DLFW:
Sort options (sort criterion, reversion, dir-first, and case-sensitivity)
are specific for each tab. Changing sort-options will not have any
affect on tabs other than the currently active one. Each new tab will
start with the default sort-options.


DLFW:
Sort options are also tab-specific in Ranger. I missed this feature many times, for example when sorting stuff from my Downloads-directory, which I mostly want sorted by ctime/mtime, to other directories which I ususally want to have sorted by file name in another tab.

If this gets merged, we could also make `show_hidden` tab-specific.

If joshuto gets other file-view specific options in the future (maybe a generic "file-name-filter"), they should then also be tab-specific.

kamiyaa (approved):

kamiyaa:
LGTM Thanks!

PR #192, opened 2022-08-15
Add "Visual-Mode"

DLFW:
This adds a "visual-mode" file selection feature where a range-selection follows the file-cursor.

Description of usage is added to the docs.

Also, the "normal" and the new "visual-mode-selection" are now preserved
when a `DirList` is reloaded.

Wrap-up of changes:
* Add command `toggle_visual`, mapped to `V`
* Add command `escape`, mapped to `ESCAPE`
* Add style `[visual_mode_selection]` for file entries which are
  "temporarily" selected by visual-mode
* For `JoshutoDirEntry`, the attribute `selected` has been renamed to
  `permanent_selected`, and a second selection-attribute
  `visual_mode_selected` has been added. "Setters" and "getters" have
  been adapted/added accordingly. The former "getter" for the `selecetd`
  attribute still exists and returns `True` for an entry which is
  "permanant selected" _or_ "visual-mode selected". So any higher logic
  which acts on selected files does not need to care about "how" and
  entry is selected.
* Each `JoshutoDirList` has an optional index of the file where
  visual-mode has been entered and re-calculates the "visual-mode
  selecetd" status for each entry any time the cursor-index changes.
* The footer has been extended so it shows a "VIS" marker when the user
  is in visual-mode.

This implementation of visual-mode is a bit different from the ranger
one, where the visual-selection is turned into a "normal selection" when
a command (like `copy-files`) is issued. This implementation keeps
both selections separate until the user toggles back to "normal mode".
Only then the visual-selection is taken over to the "normal selection".
The user also can withdraw the visual-selection with `escape`.

The `escape` command may be used also for other "reset"-actions in the
future.


DLFW:
Uhh, some Rust version problem, I guess. Will check that some other time...

src/history.rs
```diff
@@ -164,6 +166,19 @@ pub fn create_dirlist_with_history(
         }
     }
 
+    // preserve selection status of entries on reload
```
kamiyaa, line 169: This is awesome!

src/key_command/constants.rs
```diff
@@ -77,6 +78,7 @@ cmd_constants![
     (CMD_ZOXIDE, "z"),
     (CMD_ZOXIDE_INTERACTIVE, "zi"),
     (CMD_FLAT, "flat"),
+    (CMD_ESCAPE, "escape"),
```
kamiyaa, line 81: Calling it escape seems a bit ambiguous, is there any reason for this? ie. did ranger call it escape?

I'm thinking it might be better to call it `mode_exit` or `mode_escape`?

wdyt?

kamiyaa (commented):
Looks good overall, just a minor comment

src/key_command/constants.rs
DLFW: Hi!
No, ranger does not call it “escape”. Ranger uses this mapping: `map <esc> change_mode normal`. It seems (at least in my old `rc.conf`), that `change_mode` is only used here. For entering visual, Ranger uses `map V  toggle_visual_mode`.

Anyway, my idea for this `escape` command was that it can be used to escape from anywhere. I did not want to be so greedy and use Escape just for exiting visual mode. :smile:  Joshuto cannot handle key-mappings per mode (and I don't think it's necessary), so other features which also activate some kind of “temporary state” could also use the escape-command to “go back to normal”. If we name it something like `mode_escape`, it would be counter-intuitive to use the same command for resetting other things.

Anyway, if you prefer to name it (and use it) only for exiting visual mode, I'm totally ok with that.

kamiyaa: Okay, in that case, we can keep `escape`.

Maybe in the future, if different modes and different keymappings per mode is added, we can change it.

kamiyaa (approved):

DLFW:
Thanks! ;)

PR #194, opened 2022-08-31
use hashmap and uuid to store tabs

kamiyaa:
This is preliminary changes in order to track preview threads and progress.

The current setup is we just kick off a new thread to load
the given directory whenever we see the directory content
does not exist in history.
We don't track these threads or which tab these requests
came from. When the result is returned, we just assign it
to the current tab, instead of the tab that actually
initiated the request.

By adding uuid, we can now track which tab requested the
preview and assign it accordingly.
This will also allow us to track the status of the preview,
so we can display to the user a loading state,
when a directory is taking longer than usual to load.
This will also solve the problem of kicking off multiple
threads to read the same directory for the same tab.
Now these threads can be stored and tracked.

 - side: fix reload not honouring tab sort options @DLFW 
   - use tab specific options whenever we need to reload stuff

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -392,6 +392,7 @@ dependencies = [
  "unicode-segmentation",
  "unicode-width",
  "users",
+ "uuid",
  "walkdir",
  "whoami",
  "xdg",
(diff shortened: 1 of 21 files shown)


DLFW:
Excellent! Thanks!

Issue #202, opened 2022-09-15
Mode = "hsplit" not opening location


psylopneunonymous:
I just installed from cargo on void linux using `sudo cargo install --path=. --force --root=/usr/local` and copied the whole config directory into `~/.config/joshuto/`
 
I'm coming from Midnight Commander so the first thing I changed was line 11 to `mode = "hsplit"`, assuming it would give me a two-paned layout like mc. First, am I barking up the right tree to begin with or is "hsplit" something else entirely? 

Running the program with `mode = "hsplit"` in the joshtuo.toml seems to cause joshuto to crash instantly, without any error messages. Everything else is pasted from the default config file, and changing line 11 back to default makes everything work as expected. Is joshuto not crashing, but instead actually doing something I don't yet understand?

Vinegret43:
> First, am I barking up the right tree to begin with or is "hsplit" something else entirely?

Hsplit does exactly what you need - makes Joshuto act like MC by displaying two independent splits

> Is joshuto not crashing, but instead actually doing something I don't yet understand?

No, it's crashing because of a bug. All further information is for someone who wants to fix this. Made some research, it happens on [this line](https://github.com/kamiyaa/joshuto/blob/main/src/ui/views/tui_hsplit_view.rs#L141). From what i can tell, hsplit expects joshuto to have two different tabs opened - it displays the first tab on the left, and the second one on the right. Before [these changes](https://github.com/kamiyaa/joshuto/commit/0ca4ffd657f1a2884ca8422be8a78baa42f00de4#diff-b189db01c2863e3385cd86ab3f5157c9f0c586fc11ca507e6c8fb338271e8d8e) were made, this code used to just get an `Option<&JoshutoTab>` by its index, but now we get tabs by their UUIDs, so firstly it obtains UUID of a tab by its index, like this: `let other_tab_id = tab_context.tab_order[other_tab_index]`, which causes Joshuto to crash

PR #203, opened 2022-09-16
fix bug when HSplit view caused Joshuto to crash

Vinegret43:
Closes: #202 

Setting `mode` to `"hsplit"` in **joshuto.toml** caused Joshuto to crash with "index out of bounds" error. This pull makes it function like before - if second tab is not opened, only tab on the left will be displayed


kamiyaa (approved):

kamiyaa:
LGTM Thanks!

PR #205, opened 2022-09-25
add `filter` command

krivahtoo:


kamiyaa (approved):
LGTM!
I think for the future, we can display the filter setting in the bottom bar so its more clear to the user whether the folder is filtered or if the folder is just empty.
But we can add that in another PR.

DLFW:
> for the future, we can display the filter setting in the bottom bar

Agree. Ranger also does that and it really makes sense. I would even propose to also show if hidden files are shown or not (even though ranger doesn't do that, but it's also kind of a filter). It tricked me a few times already. :blush: 

BTW, A little documentation would have been nice. Just for people having a chance to see that this feature exists. 
But *Great Feature*! :partying_face:  Thanks a lot, @krivahtoo!

I came across two bugs though.

* Say you're in a dir which contains a file, matching `x`, but no file matching `xy`. Do  `:filter xy` and then “backspace” the `y` away. Then the file matching `x` won't be shown.
  Example: go into Joshuto's repo and do a `:filter mdd` and then remove the second `d`. The file `README.md` won't be shown. If I remove the first `d` also and then type it again, `README.md` is shown again.
* Apply a filer and then trigger an update, eg. by pressing `zh` to toggle hidden files. The filter will now also be applied to the parent and child panel. The filter should only be applied to the middle panel.

Also, unlike ranger, here the filter is applied for a tab. Ranger applies it per directory. I would need to think if this might just be a matter of taste, but keeping a filter applied when changing dirs feels strange for me and might lower the value of this feature when the user needs to always reset the filter when he wants to navigate to a different place in between.

krivahtoo:
@kamiyaa @DLFW Thanks for the reply.

> Say you're in a dir which contains a file...

I didn't notice this when testing. This is probably due to interactive commands not executing on 'backspace', since it displays the correctly filtered list on "enter".

> Also, unlike ranger, here the filter is applied for a tab. Ranger applies it per directory...

Agreed. It should be directory specific, also filtered directory should not reset after refresh. I was thinking of adding options that are dirlist specific for this command and `:flat` command (not sure how yet, maybe under tab options). It should fix the second bug.

PR #207, opened 2022-10-16
Add ranger-like mouse control, incl. opening files

DLFW:
This enhances the mouse control and makes it similar to the behavior of ranger.

* Entries in the “child-panel” can also “be clicked” now.
* Left-clicking entries in the parent or child-panel moves the containing dir to the middle panel.
* Right-clicking entries in the parent or middle panel opens the entry.

This feature does not work properly in `hsplit`-mode.

A short user documentation has been added to `misc.md`.


kamiyaa (approved):
LGTM!

PR #209, opened 2022-10-21
Rework filter and flat commands to be dirlist specific

krivahtoo:
- `filter` and `flat` commands are now specific to directory list and refreshing wouldn't reset the commands.
- Pressing "backspace" now updates on interactive commands.


kamiyaa (approved):
LGTM!

PR #217, opened 2022-10-24
Enhace keymap docs (incl. shell & toggle_visual)

DLFW:
* Better description for `spawn` and `shell`, now mentioning `%s` and with “%s-examples”.
* Describe `toggle_visual` and `escape`.
* Few random, tiny enhancements


kamiyaa (approved):

kamiyaa:
Thank you!

DLFW:
Hi @kamiyaa, my pleasure. ;) However, something seems to be strange here; the PR says its merged, but the commit does not appear in the commit history and it seems the changes itself are not on `main`. Was there maybe some force-push or something which reverted the commit?

PR #226, opened 2022-11-21
config option for close_tab quit action

pantosaur:
I added a configuration option for what happens when you exit joshuto through the close_tab command rather than the quit command. The default is noop, as the regular behavior, but I like to have it do the "output directory" action.

### DIFF
--- a/src/commands/quit.rs
+++ b/src/commands/quit.rs
@@ -22,6 +22,13 @@ impl QuitAction {
             Self::OutputSelectedFiles => 102,
         }
     }
+    pub fn from_str(s: &str) -> Option<Self> {
+        match s {
+            "noop" => Some(Self::Noop),
+            "outputdir" => Some(Self::OutputCurrentDirectory),
+            _ => None,
+        }
+    }
 }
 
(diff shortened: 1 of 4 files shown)


kamiyaa:
What is the use case for this?

pantosaur:
I don't really prefer to use a separate command to close a tab and to quit. And since the close_tab and quit commands are two different commands, there was no way to have an action be taken (such as output dir) upon quitting without the quit command. That's why i think it would be good to add a configuration option for the default action for every way to quit the UI. Thinking about it, I should've probably gone with an option that sets a default quit action (default noop) for QuitAction within an allowed set of options, and have a default implementation for QuitAction that reads that config option. Rather that doing something specific on close_tab. This way we would also have the same usual arguments for the quit command, but with a default "no argument" behavior that is configured with the option

pantosaur:
I can do it that way and make another PR if you accept

kamiyaa:
Yeah, I think I like the options in config better than close_tab.
Mostly because it seems kind of confusing to have something like `close_tab --output-dir` or something, but it behaves as the same as regular `close_tab` when multiple tabs are involved

PR #228, opened 2022-11-23
quit_option configuration option

pantosaur:
follow up on #226 

quit::quit_with_action now takes an `Option<QuitAction>` rather than a QuitAction. If None, it will get the value from the config, which defaults to noop, unless the quit_action config key is set. close_tab uses None to default to config, and all other calls to quit_with_action have their `Option<QuitAction>` parameter unchanged.
 
Command::Quit now encapsulates an `Option<QuitAction>` rather than a QuitAction. This allows to have Quit(None) as a default when using the quit command without argument. "--noop" is now a valid argument to noop explicitly

### DIFF
--- a/docs/command_arguments.md
+++ b/docs/command_arguments.md
@@ -11,8 +11,8 @@ Joshuto supports the following options from the command line:
 
  - `-h` `--help`: shows help menu
 
- - `--change-directory`: sets the default quit behavior to change directory
-   instead of noop
+ - `--change-directory`: sets the quit behavior to change directory instead of
+   noop when closing the last tab
 
  - `--output-file <output-file>`: tells joshuto to output data to `<output-file>`.
    - This is usually used so programs can know how to behave after joshuto exits.
--- a/docs/command_arguments.md
+++ b/docs/command_arguments.md
@@ -11,6 +11,9 @@ Joshuto supports the following options from the command line:
 
  - `-h` `--help`: shows help menu
 
+ - `--change-directory`: sets the default quit behavior to change directory
+   instead of noop
+
  - `--output-file <output-file>`: tells joshuto to output data to `<output-file>`.
    - This is usually used so programs can know how to behave after joshuto exits.
    - For example, cd into joshuto's current directory on quit
--- a/src/commands/open_file.rs
+++ b/src/commands/open_file.rs
@@ -151,7 +151,7 @@ pub fn open(context: &mut AppContext, backend: &mut AppBackend) -> JoshutoResult
         }
         Some(entry) => {
             if context.args.file_chooser {
-                return quit::quit_with_action(context, quit::QuitAction::OutputSelectedFiles);
+                return quit::quit_with_action(context, Some(quit::QuitAction::OutputSelectedFiles));
             }
 
             let paths = curr_list.map_or_else(Vec::new, |s| s.iter_selected().cloned().collect());
--- a/src/commands/open_file.rs
+++ b/src/commands/open_file.rs
@@ -151,7 +151,7 @@ pub fn open(context: &mut AppContext, backend: &mut AppBackend) -> JoshutoResult
         }
         Some(entry) => {
             if context.args.file_chooser {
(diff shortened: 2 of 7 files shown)


kamiyaa:
Looking at this more, I think a better solution is to just have a command line argument that allows for this behaviour. It's more straightforward and less layers of quit logic.

ie.
```bash
~$ joshuto --change-directory
```

similar to `joshuto --file-chooser`

What do you think?

pantosaur:
Personally, I like to have every option configurable through the config file, since you can keep track of only one file to be your configuration in a portable/shareable way, and use the CLI arguments as a way to override those options if needed.

The config option code is fully contained within quit_raw.rs and quit_option.rs, which are modeled after some other config code and which can easily be removed if no longer needed in the program. The only new thing is the requirement to generate a QuitAction from a configuration option, but would also be needed in the case of a CLI argument. In this case this is QuitAction::from_str, which i made in a similar way as TabHomePage::from_str

The part about `Option<QuitAction>` feels like the most elegant way to tell quit_with_action() what we want to do, i.e. chose something according to what the program requires (Some) or pick from a user config where the program doesn't require something specific (None). But if we want to stick with QuitAction directly, we can always go with an "if" clause to check if the config/cli option is set in the specific place where the option is allowed.

Overall, I think this way is pretty good, but I don't mind if we end up with something else

kamiyaa:
I'm still a bit unsure about this.

I feel like its adding too much complexity to the quit concept.
Having the ability to specify the quit action via 3 methods seems more complicated than it needs to be.

1. Via `quit` command
2. Via command line argument
3. Via config option

> Personally, I like to have every option configurable through the config file, since you can keep track of only one file to be your configuration in a portable/shareable way, and use the CLI arguments as a way to override those options if needed.

I think having an alias for your shell config should suffice. `alias joshuto="joshuto --change-directory --output-file abc.txt"`

pantosaur:
I changed it to be in line with what you describe. I deleted everything related to the toml config option, and added a cli parameter instead. I kept the `Option<QuitAction>` in the signature of quit_with_action, since it allows for any code that calls quit_with_action with None to default to something based on context.args, rather than put the same logic in possibly multiple calls to quit_with_action. Also, noop can be used as an explicit action, i.e. in quit --noop. All things considered, I like it this way.

pantosaur:
Is there still consideration for this? I can change it so to look at `context.args` from within `close_tab` only which would leave `quit_with_action` and everything else untouched, and discard the idea of having a default quit action.

I had in mind that `None` could be used by any code that wants to quit without explicitly selecting an action and default to something configurable instead, but that might never be necessary.

kamiyaa:
Will take a look when I have time! Just been a bit busy atm

kamiyaa:
> Is there still consideration for this? I can change it so to look at `context.args` from within `close_tab` only which would leave `quit_with_action` and everything else untouched, and discard the idea of having a default quit action.
> 
> I had in mind that `None` could be used by any code that wants to quit without explicitly selecting an action and default to something configurable instead, but that might never be necessary.

Yeah, I still think this is the best approach right now:

> I think having an alias for your shell config should suffice. alias joshuto="joshuto --change-directory --output-file abc.txt"

If more complex use cases show up in the future, we can potentially explore options for configuring the default quit action

pantosaur:
it should be good now, as simple as it can be :)

kamiyaa:
Thanks! :)

Sonico98:
Thanks both of you for your work!

I'm running joshuto using the function described on [the keymap docs](https://github.com/kamiyaa/joshuto/blob/main/docs/configuration/keymap.toml.md#quit-quit-joshuto), and modified it to also pass `--change-directory` to joshuto. It does change directories if I exit joshuto by running the `close_tab` command, but not if I use the `quit` command. Instead, I have to use `quit --output-current-directory`.
Is this the intended behaviour? I thought passing `--change-directory` as a command line argument would make joshuto _always_ change directory, no matter in which way I quit the program (in fact, I thought the function described in the docs would no longer be necessary). Or maybe it's just me getting confused :P

PR #232, opened 2022-12-03
Display filter and flat status in footer

krivahtoo:
Added footer indicators for filter and flat commands, to display current value if set.


kamiyaa (approved):

kamiyaa:
LGTM, thanks!

PR #234, opened 2022-12-03
Bulk-rename works with only current file

DLFW:
If no file is selected, bulk-rename will use the "current" file (the one with the cursor) for renaming.

Adds method `JoshutoDirList.selected_or_current -> Vec<&JoshutoDirEntry>` which can be used anywhere to get this a collection of selected files _or_ the current one if none is selected.


DLFW:
Before, bulk-rename just did nothing when called without having files selected.

This is convenient when a user wants to rename only a single file with a proper text editor.

kamiyaa (approved):
LGTM

PR #249, opened 2023-01-24
Add option to override deleting to trash

krivahtoo:
Add the `--permanently` option to the `delete_files` command to delete files permanently when `use_trash` is `true`.


kamiyaa:
LGTM! I will need to get around to fixing the trash feature because it doesn't work sometimes on mounted drives

PR #253, opened 2023-02-04
Help search also inside description column

yanshay:
I'm new to joshuto so use the Help view.
Currently when searching in the help using '/' it searches only the key and command columns. I actually need the search to search in the 'Description' column. 
So I modified the search code to search also in the Description column and do a case insensitive search.


kamiyaa (approved):
LGTM! Thanks!

kamiyaa:
Can you run `cargo fmt`? Thanks!

yanshay:
done

PR #277, opened 2023-04-12
autocomplete: assume selection when there's only one candidate

adtac:
This makes autocomplete behave like how it works in ranger. To test before and after, create a directory long directory chain `$HOME/foo/bar/baz`, type `:cd fo` and press Tab several times. Previously, joshuto would stop at `foo`, but now the first Tab will autocomplete `foo/`, the second will autocomplete `foo/bar/`, and so on.

I don't know any Rust, but the change seems to work. Feel free to cherry pick the commit and make changes to make it more idiomatic/cleaner/whatever.

The PR also includes a couple of other trivial improvements.

### DIFF
--- a/docs/image_previews.md
+++ b/docs/image_previews.md
@@ -265,7 +265,7 @@ case "$mimetype" in
 esac
 ```
 
-`~/.config/joshuto/on_preview_removed.sh`:
+`~/.config/joshuto/on_preview_removed`:
 ```shell
 #!/usr/bin/env bash
 
--- a/src/key_command/impl_from_str.rs
+++ b/src/key_command/impl_from_str.rs
@@ -248,7 +248,6 @@ impl std::str::FromStr for Command {
         } else if command == CMD_DELETE_FILES {
             let (mut permanently, mut background) = (false, false);
             for arg in arg.split_whitespace() {
-                eprintln!("arg: {:?}", arg);
                 match arg {
                     "--background=true" => background = true,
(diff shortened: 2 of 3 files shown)


src/key_command/impl_from_str.rs
```diff
@@ -248,7 +248,6 @@ impl std::str::FromStr for Command {
         } else if command == CMD_DELETE_FILES {
             let (mut permanently, mut background) = (false, false);
             for arg in arg.split_whitespace() {
-                eprintln!("arg: {:?}", arg);
```
kamiyaa, line 251: Oops, thanks for catching this!

docs/image_previews.md
```diff
@@ -265,7 +265,7 @@ case "$mimetype" in
 esac
 ```
 
-`~/.config/joshuto/on_preview_removed.sh`:
```
kamiyaa, line 268: Is there any reason for this change?

kamiyaa (commented):
LGTM overall.

Just a few minor comments

docs/image_previews.md
adtac: Both references to this script above ("Configuring Hook Scripts" and "The Hook Scripts") don't have the .sh. I blindly copy-pasted those instructions, but with the kitty modifications, but it took me a couple minutes to identify the typo.

scutuatua-crypto (approved):

scutuatua-crypto (approved):

Issue #282, opened 2023-04-17
No text previews


nat-418:
I don't understand how to make text previews work like with `lf`, `ranger`, etc. I don't care about image previews, I just want to see the file contents in the far right pane. I've looked through the docs, issues, etc. and I am not sure if this is supposed to be automatic and something is wrong with my install or if I need to make a wrapper script or configuration change from the defaults to enable the feature.

DLFW:
Hm, yeah, the documentation is lacking there...
Have you discovered the [example config dir](https://github.com/kamiyaa/joshuto/tree/main/config)?
It contains an example-script (`preview_file.sh`) which generates textual previews and the `joshuto.toml` there shows how to set that script as the “preview generator”.

Try to copy that preview script and add the corresponding entry into your `joshuto.toml`. (Within _your_ `~/.config/joshuto`, of course.)

Actually, this is the exact same concept as in _ranger_. In fact, the preview-script of _joshuto_ has been “forked” from _ranger_'s preview script. _Joshuto_'s preview script may still have residuals or even some dead code from that heritage. Not sure. I tried to clean it up some time ago.

nat-418:
Yes, I did look in there and copy the file and search the issues etc. The issue is that it doesn't work.

kamiyaa:
Hey, thanks for the issue.
I'll make a writeup/documentation on the preview aspect of joshuto :+1:

nat-418:
Here's what I did:

1. I copied the example joshuto.toml to ~/.config/joshuto/joshuto.toml
2. I copied the example preview_file.sh to ~/.config/joshuto/preview_file.sh, marked it executable, and put it on my $PATH.
3. I did the same for the example wrapper script.

DLFW:
> 4\. I added `use_preview_script = true` to the bottom of my ~/.config/joshuto/joshuto.toml file.

Where is that coming from?
You need some
```
[preview]
...
preview_script = "~/.config/joshuto/preview.sh"
...
```


> 3\. I did the same for the example wrapper script.

Not sure what you mean. The example wrapper for the image-previews? If so: that's not necessary for the text preview (but it doesn't hurt either).

nat-418:
> Where is that coming from?

From reading `preview_file.sh`:

```
#!/usr/bin/env bash

set -o noclobber -o noglob -o nounset -o pipefail
IFS=$'\n'

## If the option `use_preview_script` is set to `true`,
## then this script will be called and its output will be displayed in ranger.
[...]

PR #285, opened 2023-04-23
Cleanup preview.sh template

DLFW:
Removed ranger residuals and dead code from the preview.sh script and the logic that invokes it.

### DIFF
--- a/config/preview_file.sh
+++ b/config/preview_file.sh
@@ -1,8 +1,5 @@
 #!/usr/bin/env bash
 
-set -o noclobber -o noglob -o nounset -o pipefail
-IFS=$'\n'
-
 ## This script is a template script for creating textual file previews in Joshuto.
 ##
 ## Copy this script to your Joshuto configuration directory and refer to this
@@ -12,8 +9,8 @@ IFS=$'\n'
 ## ```
 ## Joshuto will call this script for each file when first hovered by the cursor.
 ## If this script returns with an exit code 0, the stdout of this script will be 
-## the file's preview text in Joshuto's right panel. The preview text will be cached
-## by Joshuto and only renewed on reload.
+## the file's preview text in Joshuto's right panel.
+## The preview text will be cached by Joshuto and only renewed on reload.
 ## ANSI color codes are supported if Joshuto is build with the `syntax_highlight`
 ## feature.
 ##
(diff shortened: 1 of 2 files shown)


DLFW:
As a user [recently pointed out](https://github.com/kamiyaa/joshuto/issues/282#issuecomment-1513803726), there was quite some misleading stuff in the `preview.sh` template script, shipped with Joshuto.

## Changes

### Intro Comment
The intro comment in the script has been rewritten. It was still describing the ranger usage.

### Exit Codes
Joshuto seems to only check for exit code 1 / != 1. So, the table of exit codes in the script's intro comment has been cut down to the values 0 and 1. This should match the Joshuto functionality. All `exit [>1]` have been replaced by `exit 0`.

### Remove --image-cache
The `--image-cache` parameter has been removed from both, the script and the Joshuto code that calls it. It was unused and was always zero. Also, the motivation was unclear to me.

## Question
I thought about removing also the  `--x-coord` and the `--y-coord` parameters. I don't see in what scenario they can be relevant for creating a text-preview. However, I first wanted to ask if that's ok.

Actually, also the `--preview-width` and the `--preview-height` parameter do not really make sense right now as Joshuto does not refresh the text-preview cache on geometry-change. (Or am I wrong here?) However, maybe it may still be helpful to have them.

## Side Note
I think it would be a good idea to change the struct
```rust
pub struct FilePreview {
    pub status: std::process::ExitStatus,
    pub output: String,
    pub index: usize,
    pub modified: time::SystemTime,
}
```
so that the status is a meaningful enumeration. If I will ever pick up a feature that introduces a new exit-code logic, I would do that....

DLFW:
Oh yeah, and I re-activated the preview for video/audio and for PDF. Not sure why these parts where commented out?

Also, the preview of epub now also uses bat for coloring the markdown translation.

kamiyaa:
> ### Intro Comment
> 
> The intro comment in the script has been rewritten. It was still describing the ranger usage.

Yeah, this was due for change 👍 

> ### Exit Codes
> 
> Joshuto seems to only check for exit code 1 / != 1. So, the table of exit codes in the script's intro comment has been cut down to the values 0 and 1. This should match the Joshuto functionality. All `exit [>1]` have been replaced by `exit 0`.

Yeah, I was thinking down the line we would implement the rest.
But I guess we can cross that bridge once it comes.

> ### Remove --image-cache
> 
> The `--image-cache` parameter has been removed from both, the script and the Joshuto code that calls it. It was unused and was always zero. Also, the motivation was unclear to me.

This was also residual from ranger.
I think the feature itself is useful, but maybe theres a better way to do it than `--image-cache`

> ## Question
> 
> I thought about removing also the `--x-coord` and the `--y-coord` parameters. I don't see in what scenario they can be relevant for creating a text-preview. However, I first wanted to ask if that's ok.

Yeah, I think it can be removed for text previews, but it is required for image previews.

> 
> Actually, also the `--preview-width` and the `--preview-height` parameter do not really make sense right now as Joshuto does not refresh the text-preview cache on geometry-change. (Or am I wrong here?) However, maybe it may still be helpful to have them.

Yes, you are correct here.

> ## Side Note
> 
> I think it would be a good idea to change the struct
> 
> ```rust
> pub struct FilePreview {
>     pub status: std::process::ExitStatus,
>     pub output: String,
>     pub index: usize,
>     pub modified: time::SystemTime,
> }
> ```
> 
> so that the status is a meaningful enumeration. If I will ever pick up a feature that introduces a new exit-code logic, I would do that....

Yeah that works 👍 

Let me know if you have any other questions.
As always thanks for the contributions!
Once the PR is ready for review, I will review it 👍

DLFW:
Hi @kamiyaa, this is ready for review, just in case it got missed. :wink:

kamiyaa:
LGTM, thanks!

Issue #289, opened 2023-05-03
move from tui to ratatui


Ragnyll:
it looks like tui, a core dependency to this project is unmaintained. see https://github.com/fdehau/tui-rs/issues/654
The maintainer and community behind the project has moved to https://github.com/tui-rs-revival/ratatui .

I think that it's in the interest of this project to move the dependency to ratatui.

Ragnyll:
I'll start looking into making this switch if you want to assign this issue to me. I imagine that i'll probably just find the version with parity to the current dependency, then look at bumping it after that. It sounds like the new maintainers are looking to do some major changes in the coming future.

kamiyaa:
Interesting! I wasn't aware of this and I'll definitely look into helping the tui community.

Yep, I can assign this to you. Thank you! ♥️

Ragnyll:
Opened the pr https://github.com/kamiyaa/joshuto/pull/290

kamiyaa:
Closed via #290

PR #290, opened 2023-05-04
change dependency from tui to ratatui

Ragnyll:
#289 
This changes the tui depedency to the ratatui package and makes the versioning less permissive since breaking changes are likely coming on minor versions

This version of ratatui (0.20.1) should have parity with tui, thought it looks like there are some type issues that i had to fix as well. 

I tested this just by hopping into joshuto and going between files and dirs, opened a file using it, and creating some tabs. If there are more manual tests i should run lemme know. 

tested on 6.3.1-arch1-1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,13 +28,13 @@ dependencies = [
 
 [[package]]
 name = "ansi-to-tui"
-version = "2.0.0"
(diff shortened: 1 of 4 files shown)


kamiyaa:
LGTM thanks!

PR #299, opened 2023-05-24
Added new command "search_fzf_rec"

Jacky-Lzx:
I'm adding a new command to let `fzf` recursively searches all files in the current directory, and make the cursor jump to the corresponding location.

https://github.com/kamiyaa/joshuto/assets/49983642/244dd0ca-3dc2-46d1-81a1-f50178cab8dc

### DIFF
--- a/src/commands/search_fzf.rs
+++ b/src/commands/search_fzf.rs
@@ -1,5 +1,6 @@
 use std::io;
 use std::io::Write;
+use std::path::Path;
 use std::process::{Command, Stdio};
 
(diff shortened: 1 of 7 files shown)


theniceboy:
How is this different than `subdir_fzf`?

Jacky-Lzx:
> How is this different than `subdir_fzf`?

Oh actually it's the same...

I failed to run `subdir_fzf` once, so I wrongly thought there was no such command...

Issue #241, opened 2022-12-14
[REQUEST] Remember tabs between sessions


Sonico98:
It would be great if joshuto could remember which tabs were opened the last session, and restore them when opening the file manager again.

DLFW:
Just an idea:
1. We add a “placeholder” `%t` for shell/spawn arguments that expands to a list of the current directories of all open tabs
2. We add a command that reads a list of directories from an external script and opens them as tabs

With that, the user can:
* Write a script that saves the current set of tabs to a certain text-file and bind it to a key-chord
* Write a script that reads from that files and bind it to another key-chord to recover the tab-set

Additionally, the user can:
* Write a script that saves the current set of tabs to a file whose name can be chosen (for example with rofi or dmenu)
* Write a script that opens dmenu or rofi where one can select one of these stored “sessions” (tab-sets) and Joshuto would open them

This second thing would also be really useful for me. There are certain directory-sets I often open at the same time.
This would allow me to prepare such dir-lists in text files and quickly choose a tab-set with rofi.

The command for opening tabs from a list could have an argument switch to choose if the current tabs shall be closed/replaced, or kept open (and the new dirs would open in additional tabs/appended).

That would also allow to use any other source of dir-lists to open a corresponding tab-set.

This solution would not include an “auto-save” when closing Joshuto though.
Maybe we could add a general feature to execute a list of commands to be executed when opening and another one to be executed when closing later on?

Opinions?

Sonico98:
It sounds useful when thinking about the rofi script for managing sessions. However, it sounds like a bit of a pain from the development side (speaking for you and kamiyaa, as I don't know Rust), and overly complicated for new users to set up (they're already finding it dificult to figure out how to implement previews).
Also IMHO the autocommands on opening and closing would be mandatory because we as humans tend to make errors, and I can see users forgetting to save their tabs every time they want to close the editor.

If you decide to go the script route, maybe it's possible to ship joshuto with the scripts in the ~/.config folder by default? That way users would have it a bit easier to set up. Also I may be able to help with the documentation.

Of course I'm just another user, so @kamiyaa's opinion is much more important.

DLFW:
> it sounds like a bit of a pain from the development side

No, I don't think that this would be a particularly hard one. But thanks for the sympathy. :wink: :heart:

> overly complicated for new users to set up

Hm, yeah. You have a point there. I might argue that this is not a very important feature that needs to work out-of-the-box, but still, you have a point there...

[...]

PR #306, opened 2023-05-25
Add a new feature: fzf bookmarks

Jacky-Lzx:
I'm implementing fzf bookmarks inspired from [ranger-fzf-marks](https://github.com/laggardkernel/ranger-fzf-marks), which is a plugin I use everyday in ranger, and I'd like to have it as well in joshuto.

It acts as follows:

- `fzf_add_mark [markname]` creates a new line in file `$HOME/.fzf-marks`, with the format `markname : current_directory`.
- `fzf_cd_mark` opens a `fzf` selection window to select a bookmark, and then `cd` to that directory in joshuto.

See it in the following video:

https://github.com/kamiyaa/joshuto/assets/49983642/e33f197c-74d9-4559-8663-c8972e54bfa3

### DIFF
--- a/docs/configuration/keymap.toml.md
+++ b/docs/configuration/keymap.toml.md
@@ -287,6 +287,10 @@ When disabling, the current “visual mode selection” is turned into normal se
 ### `add_bookmark`: adds a bookmark to the `bookmarks.toml` file
 ### `cd_bookmark`: prompts a menu of bookmarks and navigates to selected bookmark
 
+### `fzf_add_mark <name>`: create a fzf bookmark entry in file `#HOME/.fzf-marks`, with the format `<name> : <current directory path>`.
+
+### `fzf_cd_mark`: open a fzf window to let user select an entry in `#HOME/.fzf-marks`, and cd to the selected directory.
+
 ## Integration
 ### `bulk_rename`: rename all selected files
 - this will create a file inside `$TMP_DIR` (or `/tmp` if `$TMP_DIR` is not set) and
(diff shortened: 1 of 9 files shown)


Jacky-Lzx:
I don't know if you like it. Maybe I'm the only one using it everyday. I'd like to make it as a plugin in the feature.

If you like it, please give me some comments about my codes and I'll adjust them. I'm new to rust. And I'll add documents before merging.

tracy0880550287:
It's inspiring to see such high-quality open-source projects.

Jacky-Lzx:
Updated the document.

DLFW:
Hi!
Just some thoughts from my side:

IMHO, this is a pretty specific feature which would be more suited for a plugin or an external script. Joshuto has no plugin concept, but in context of #241, I [proposed](https://github.com/kamiyaa/joshuto/issues/241#issuecomment-1546851335) that we can “chain” commands for key-chords and that we can provide the output of the last spawn command as variables the following commands. If I understood this feature right, it should be doable with minimal external scripting then. 

The keymap might have stuff like this then:
```toml
keymap = [
  {
    keys = [ "f ", "z", "a" ],    # fuzzy add
      command = [
        "shell add_current_dir_to_marks.sh {dir}",     # {dir} gets substituted by the current directory of the current tab
      ]
  },
  {
    keys = [ "f", "z", "c" ],    # fuzzy change
      command = [
        "shell fuzzy-pick-from-marks.sh",
        "cd {stdout}"      # {stdout} gets substituted with the stdout of the last `shell` command
      ]
  },
]
```
The only difference would be that one has to ask for the “mark name” from the script instead of giving it as parameter in Joshuto's command line.

What do you think?

Jacky-Lzx:
Oh yes. Your solution makes more sense to me as well. I might think about it when I have spare time.

Issue #300, opened 2023-05-24
Program hangs when executing `subdir_fzf` on a large directory


Jacky-Lzx:
Problem description:

1. When executing `subdir_fzf` command in a large directory, _e.g._, `/`, the view goes to the command line, and it waits for outputs.
2. Now press `ctrl+c` will exit joshutu, instead of terminating `subdir_fzf` execution.

Why is this a problem?

- It is not a good practice to do `subdir_fzf` on such a big directory, but sometimes I may accidentally hit the key and execute `subdir_fzf`. 
- When I realize it I want to terminate the execution and return back to joshuto, but now the `ctrl+c` directly terminates joshuto.

Steps to reproduce:

1. go to `/`.
4. execute `subdir_fzf`, and the program hangs.
5. press `ctrl+c`, and joshutu exits instead of terminating `subdir_fzf` and returning back to joshutu.

Jacky-Lzx:
I think the program is waiting for the output of `FZF_DEFAULT_COMMAND` here:

https://github.com/kamiyaa/joshuto/blob/95d304fb04a87fbb2534759177f6cd27f5816143/src/commands/subdir_fzf.rs#LL17C1-L28C7

Commenting it and not redirecting stdin of `fzf` seem to fix the problem.

 I think it's fine because `fzf` will call `FZF_DEFAULT_COMMAD` by itself (I'm not sure).

kamiyaa:
Closed via https://github.com/kamiyaa/joshuto/pull/309

PR #309, opened 2023-05-27
fix: `subdir_fzf` stuck on large directory.

Jacky-Lzx:
Fixed issue #300.

`fzf` command will call `FZF_DEFAULT_COMMAND` by default when input is tty. Therefore no need to call it in joshuto.


kamiyaa (approved):
Looks good! I didn't realize fzf calls `FZF_DEFAULT_COMMAND` by default

PR #317, opened 2023-05-29
Tab theme configurable

DLFW:
Text style for tabs can be configured within
`theme.toml` with the dotted keys `tab.active` and `tab.inactive`.

The default style is the same as the former hard-coded style.

`theme.toml` has been restructured and comments have been added to give a short documentation on each style item.

The documentation (`theme.toml.md`) has been enhanced, and the excerpt of `theme.toml` with additional comments has been removed from there.

That way it should be easier for devs to keep docs and the default `theme.toml` in sync, and easier for users when the detail comments can be found in the configuration file itself.

The conversion from `AppStyle` to `tui::style::Style` is now provided as a function of `AppStyle` and a `From`-implementation for `Style` to avoid having the same factory code repeating all over.

### DIFF
--- a/docs/configuration/theme.toml.md
+++ b/docs/configuration/theme.toml.md
@@ -1,11 +1,14 @@
 # theme.toml
 
-This file is used to theme joshuto
+This file is used to theme Joshuto.
 
-## Style
+As of now, all theming items are "text style" items where one can define the
+background color and the foreground color, and enable some font attributes.
 
-Each style has the following fields:
 
+## Style options
+
+Each style allows the following fields:
 ```toml
 # background color
 bg = "light_blue"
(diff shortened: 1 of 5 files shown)


docs/configuration/theme.toml.md
```diff
@@ -1,11 +1,14 @@
 # theme.toml
 
-This file is used to theme joshuto
+This file is used to theme Joshuto.
 
-## Style
+As of now, all theming items are “text style” items where one can define the
```
kamiyaa, line 5: Nit, I would use "" instead of “”. Unless you have a reason to use those quotes.

docs/configuration/theme.toml.md
```diff
@@ -15,81 +18,46 @@ bold = false
...
-## Theme
+## Configuration Items
 
-joshuto supports theming via system file types as well as extensions
+You can find a complete list of the available configuration items with their default
+values and short explanations as comment in 
+[`theme.toml`](https://github.com/kamiyaa/joshuto/blob/main/config/theme.toml).
 
-System file types include:
+In general, `theme.toml` allows to specify the style of a few UI “widgets”,
```
kamiyaa, line 52: Same nit here xd

kamiyaa (commented):
Looks good overall, just a few minor nits. 
Will merge once conflict is resolved as well as the nits are addressed :+1:

DLFW:
Rebased to main, fixed comments, and squashed.

The reason I had used “” is simply that these are the typographically correct quotation marks. However, I understand that this might be confusing and deviate from other places as most people use the ASCII "" quotation. I just have them in my muscle memory. I need to remember to use the "ASCII quotes" while writing prose. :smiling_face_with_tear:

kamiyaa (approved):

kamiyaa:
Thanks!

PR #333, opened 2023-06-08
refactor: avoid changing constraints during PreviewFileState load

Handfish:
File previews load pretty quickly. Adjusting constraints during loading state made the file sizes flicker on the right side of my screen due to the rapidly growing and shrinking center column. This change keeps the columns a consistent size when highlighting new items making the application a bit easier on the eyes.


kamiyaa (approved):
Looks good :+1:

PR #334, opened 2023-06-10
feat: Display total size of current directory

### LINKED ISSUE #229: Display total size of current directory
I would like to somehow display the total size of a directory's content (like joshuto does with files). 
Out of curiousity, is there a reason this isn't implemented directly into joshuto already? The same feature exists for files. The argument against it is probably the performance overhead (I imagine it would be terribly slow for deeply nested directories). Maybe this could be added as an option that can be toggled on or off with a keybinding (I could look into it). You're in a parent directory of directories you want to know the size of, toggle the option, see the size, and toggle back off when you want performance again.
If you don't want this implemented directly into joshuto: The command `du -sh "$DIRECTORY"` would easliy let me do that for a single directory. But is there a way to manually (through a keybinding) call this command on a line in joshuto and somehow display the output somewhere in joshuto?

krivahtoo:
Add the option to display lazily loaded directory sizes.

The displayed size is not accurate, to get the actual size of a directory one has to navigate recursively to all its child directories.

```toml
[display]
# ...
# Options include:
# - count (use the number of items in the directory)
# - size (use the total size of items in a directory)
size_mode = "count"
# ...
```

## Screenshot
2023-06-10-215616_714x444_scrot

closes #229


DLFW:
Hi!

When I have not visited all sub-dirs, does Joshuto show the wrong directory size?

If yes, is there some way to know which sub-dirs I have visited and which are not included in the total size yet?

krivahtoo:
> When I have not visited all sub-dirs, does Joshuto show the wrong directory size?

Yes. Without visiting sub-dirs, it will only calculate the total size of files in the directory.

> If yes, is there some way to know which sub-dirs I have visited and which are not included in the total size yet?

No.

Having such a feature would be great though I am not sure how it would be implemented without traversing the directory.

DLFW:
Hm, I just have a concern that this behavior breaks user-expectations, because people either check the sizes of files or a directory fully recursive (maybe except of other file-system mounted in the sub-tree and alike). I have never seen such a “partial dir-size” calculation, and I would certainly not expect this. @fritzrehde was also asking for a `du -hs` equivalent in #229, and that will always include the complete sub-tree. I'm not sure if this PR would suit his use-case.

Furthermore, users would not have a chance to see that their expectation is wrong unless they realize that “something with the size does not look right”.  That's a dangerous combination in terms of usability, and in this case even more as a directory size might be considered in decisions about deleting, coping, or moving files.

I also don't see a use-case for this “partial dir-size” where I have to remember what sub-dirs I have visited before, but maybe I'm missing something.

Sorry, I don't want to interfere in your MR, and I'm really a fan of all your Joshuto contributions, but I have some serious concerns about this one.

krivahtoo:
Okay, I see what you mean. Yeah, this will probably break user expectations.

Thanks for your feedback.

I think the solution would be to calculate but only show the size when all the files/dirs have been calculated and/or add a command to traverse the directory. This way the users would see the directory size only when it has been fully calculated.

Issue #338, opened 2023-06-18
[Feature request] Plugin system (custom_commands) and custom search command


Dzordzu:
Hi it would be nice to be able to execute custom search command. For example currently it's not possible to have rg + fzf search. Ideally it may look like

keymap.toml
```toml
[default_view]

keymap = [
  { keys = ["x"], command = "custom_serach_interactive rg -n -H --color=never '%text' %s|  fzf --ansi --preview 'bat -n $(echo {} | cut -d \":\"-f1)' | cut -d ':' -f1" },
  { keys = ["X"], command = "custom_search rg -l %text | tail -n 1  },
]
```

First one should allow to open interactive window (in this example fzf) the second just go to the output file

Dzordzu:
Started working on implementation on this. Well it's a little more complicated. We may want to specify a new config handling user defined commands. I suggest creating new key in the joshuto.toml named `custom_commands`. It may look like this

joshuto.toml
```toml
custom_commands = [
   { name = "rgfzf", command = "rg -n -H --color=never '%text' %s|  fzf --ansi --preview 'bat -n $(echo {} | cut -d \":\"-f1)' | cut -d ':' -f1" }
]
```

keymap.toml
```toml
[default_view]

keymap = [
  { keys = ["X"], command = ":custom_search rgfzf"  },
]
```

superiums:
i thinks this may be restructed more flexable:

custom actions is not the point, the point was communication between commands and joshuto.

1. there are 3 ways of shell command:
- background shell command, which NOT live joshuto, display at the bottom.. (like `:shell tar -xzf aaa.tgz` does)
- inline shell command, which do NOT live joshotu, display at the bottom and pie result back to joshotu. (like / `search` cmd does)
- interactive shell command, which live joshotu and pipe result back to joshotu. (like `search_fzf` does.)

2. we analyze these needs, the custom search command was the 3rd kind.
so we could make:
- two types of UI, inline and interactive one.
- two types of result pipe: pipe back and not.

in this way, not only custom search could be done well, but also other custom actions should be work very well.

by the way, we may consider another powerful thing:
- pipe out : pipe out the selection or current dir/file to the command. (seems this has been done by %s, but maybe more)

sth more:
- cusor position in shell: just like `!` and `@` do in ranger. 
- result wait: just like `shell -w ` do in ranger.

Dzordzu:
[...]

PR #339, opened 2023-06-18
Added custom_commands and custom_search

Dzordzu:
Implemented #338

### DIFF
--- a/src/commands/change_directory.rs
+++ b/src/commands/change_directory.rs
@@ -4,10 +4,11 @@ use crate::commands::reload;
 use crate::context::AppContext;
 use crate::error::AppResult;
 use crate::history::DirectoryHistory;
+use crate::util::cwd;
 
 // ChangeDirectory command
 pub fn cd(path: &path::Path, context: &mut AppContext) -> std::io::Result<()> {
-    std::env::set_current_dir(path)?;
+    cwd::set_current_dir(path)?;
     context.tab_context_mut().curr_tab_mut().set_cwd(path);
     Ok(())
 }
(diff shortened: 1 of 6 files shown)


Dzordzu:
Demo:
joshuto_demo

Dzordzu:
@kamiyaa can you look at it?

kamiyaa:
> @kamiyaa can you look at it?

Yep! Sorry, I've just been busy :x

kamiyaa:
Looks good so far.

Is there a reason for putting it inside `joshuto.toml`?
I think it would make more sense to keep inside `keymap.toml`
But on the other hand, that might make `keymap.toml` more complicated than it needs to be

Also, if you could add documentation for the new commands inside `docs/` either in this PR, or a followup PR, that would be great!

Dzordzu:
> Is there a reason for putting it inside `joshuto.tom`l

Yep. It may be later used as the simple plugin system (not only for commands, but for example for another window)

> I think it would make more sense to keep inside `keymap.toml`

I also thought about it, but it is not a keymap, but a more general component

> But on the other hand, that might make `keymap.toml` more complicated than it needs to be

Yep

> Also, if you could add documentation for the new commands inside `docs/ `

Oh. I forgot to include it! I will create a commit with docs in the following days :)

Beethoven-n:
what're we waiting for? everything looks good from here

Dzordzu:
I need to resolve conflicts, add more documentation. Furthermore I'm required to find a spare time and will to contribute (a little personal issues, sorry for that!)

kamiyaa:
Just ping me when its ready!

Dzordzu:
I messed it so badly, that I needed to open another PR. Working at nights is really funny!

PR #349, opened 2023-06-25
Migrate termion lib from v1 to v2. Mouse feature turn on by default

asim215:
Small changes to migrate from termion v1 to v2. 
Small refactor of backend.rs for simplification and clarity.
Mouse feature turn on by default.


kamiyaa:
Looks good! Thanks!

PR #350, opened 2023-06-26
Fix incorrect sort method field in example config 

rqdmap:
When I follow the default config files in `/config/joshuto.toml`, I find the field [method](https://github.com/kamiyaa/joshuto/blob/6cb151ea4b58d006cfc52f54d52be05110bd5b90/config/joshuto.toml#L27C19-L27C19) which decides how to sort items doesn't work. It's because the default name of it in source code is [sort_method](https://github.com/kamiyaa/joshuto/blob/6cb151ea4b58d006cfc52f54d52be05110bd5b90/src/config/general/sort_raw.rs#L20C8-L20C8).

### DIFF
--- a/config/joshuto.toml
+++ b/config/joshuto.toml
@@ -23,8 +23,8 @@ tilde_in_titlebar = true
 line_number_style = "none"
 
 [display.sort]
-# lexical, mtime, natural
-method = "natural"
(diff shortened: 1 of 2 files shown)


kamiyaa (approved):
LGTM! Thanks!

PR #351, opened 2023-06-28
display the full path after the CD command with the TAB key

macdylan:
Display the full path like:
`:cd ~/Pictures/Photo\ Booth\ Library`

### DIFF
--- a/src/ui/views/tui_textfield.rs
+++ b/src/ui/views/tui_textfield.rs
@@ -313,16 +313,18 @@ fn autocomplete(
     if let Some(ref mut ct) = completion_tracker {
         ct.index = if reversed {
             ct.index.checked_sub(1).unwrap_or(ct.candidates.len() - 1)
-        } else if ct.index + 1 < ct.candidates.len() {
-            ct.index + 1
         } else {
-            ct.index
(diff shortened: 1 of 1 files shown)


src/ui/views/tui_textfield.rs
```diff
@@ -313,24 +313,26 @@ fn autocomplete(
     if let Some(ref mut ct) = completion_tracker {
         ct.index = if reversed {
             ct.index.checked_sub(1).unwrap_or(ct.candidates.len() - 1)
-        } else if ct.index + 1 < ct.candidates.len() {
-            ct.index + 1
         } else {
-            ct.index
+            (ct.index + 1) % ct.candidates.len()
```
kamiyaa, line 317: I prefer the autocomplete suggestions does not loop itself.
If a user would like to go back on a suggestion, one can use reverse order or restart the autocomplete by typing something.

kamiyaa (commented):
Overall, looks good. Make sure to format the code after as well :+1:

scutuatua-crypto (approved):

scutuatua-crypto (approved):

PR #363, opened 2023-07-10
Fix docs for combined text and image preview

DLFW:
Fixed a “bug” in the recipe for combining text and image previews. Images at a location whose path contains a space were not shown with the calculated y-offset.


kamiyaa (approved):
LGTM!

PR #370, opened 2023-07-17
feat: Add support for command aliases

Akmadan23:
I think it can be handy to have the ability to set custom command aliases (such as `:q` aliased to `:quit`), so I added a really simple implementation with a `HashMap`, which works fine for this task. For configuration I chose to put it in the main `joshuto.toml` because it's optional and if specified, the list isn't likely to be really long.
Let me know if it's a welcome feature or if there's something you would do differently.

### DIFF
--- a/docs/configuration/joshuto.toml.md
+++ b/docs/configuration/joshuto.toml.md
@@ -74,4 +74,9 @@ directories_first = true
 
 # sort in reverse
 reverse = false
+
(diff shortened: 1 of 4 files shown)


kamiyaa:
Looks good, just minor styling.
After its fixed I can merge it 👍

Akmadan23:
I wasn't aware that cargo-fmt cares about imports order, from now on I'll always double check

PR #382, opened 2023-07-29
Update README.md

pato-pan:
docs/image_previews.md doesn't exist and leads to a 404 page. Perhaps a link to the folder or the readme of the folder would be more appropriate?


kamiyaa:
Thanks!

Issue #379, opened 2023-07-27
[FR] Use system timezone for modified time instead of UTC


xfzv:
```console
❯  joshuto -v
joshuto-0.9.5
```

Unless I'm mistaken, the files modified time is based on UTC. If the system timezone is not UTC, then the data isn't relevant.

It would be nice if joshuto could use the system timezone instead, like [lf](https://github.com/gokcehan/lf) or [xplr](https://github.com/sayanarijit/xplr) for instance.

kamiyaa:
Closing because PR is merged

PR #383, opened 2023-07-30
feat: Time will display with local timezone

Lzzzzzt:
the files modified time will display with local timezone.

fix the issue #379 

## Screenshot
image


DLFW:
Nice! That was also bugging me since ever! 
Gave it a quick test and works fine, also in the “mtime linemodes”. 

Would have make sense to use `main` branch as target though.

Lzzzzzt:
> Nice! That was also bugging me since ever! Gave it a quick test and works fine, also in the “mtime linemodes”.
> 
> Would have make sense to use `main` branch as target though.

Should I close this pr a.nd create an another pr with main branch? I haven't noticed that the dev branch hasn't update for long time

DLFW:
Hi! I'm  not the maintainer, btw. So, such decisions are not on me. 🙃

However, I think it's not necessary to recreate the PR. I don't have a PR open and can't check it myself, but [it looks like it's easily possible](https://stackoverflow.com/questions/24159036/how-to-modify-a-pull-request-on-github-to-change-target-branch-to-merge-into#24159161) for the requester. (Maybe the maintainer can also change it upon merge...)

PR #384, opened 2023-07-30
feat: mtime will display with local timezone

Lzzzzzt:
the files modified time will display with local timezone.

fix the issue #379 

change dev to main(#383 )


kamiyaa:
Yeah, this looks good. I didn't really bother with timezones at the time.
But a very welcomed change!

PR #389, opened 2023-08-02
fix #388

### LINKED ISSUE #388: Bug on `:cd ../*`
I noticed that when running a `cd` to a relative path that starts with `../` the directory change works fine but there is something that will be messed up. Below are a few examples.

## Example 1
Initial state:
immagine
### After running `:cd ../..`
Expected behavior:
immagine
Actual behavior:
immagine

## Example 2
Initial state:
immagine
### After running `:cd ../dir3_alt`
Expected behavior:
immagine
Actual behavior:
immagine
___

I already found a fix for this, I'll make a PR right after this issue.

Akmadan23:
This fixes the bug related to `:cd ../*` described in #388.

### DIFF
--- a/src/commands/change_directory.rs
+++ b/src/commands/change_directory.rs
@@ -12,10 +12,15 @@ pub fn cd(path: &path::Path, context: &mut AppContext) -> std::io::Result<()> {
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


kamiyaa (approved):
LGTM 👍

PR #400, opened 2023-08-13
feat: add `noconfirm` option to `delete_files` command

Akmadan23:
Sometimes having to confirm every file deletion can be annoying, especially if `use_trash` is enabled, so I added this option to skip the prompt.
In `commands/delete_files.rs` I moved the prompt part in a separate function so that it's easier to skip, and left in the `delete_files` function only the part that actually deletes the files.


kamiyaa (approved):

kamiyaa:
LGTM, Thanks!

PR #411, opened 2023-08-23
feat: regex support

ctj12461:
Search files using regex pattern.


kamiyaa (approved):
Looks good! Thanks for the feature!

PR #412, opened 2023-08-24
Fix: macos build on automated-build

kennycallado:
This PR addresses the macOS build failure in the automated build process. The issue was caused by [deprecated](https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/) versions.

After merging this pull request, note that the macOS binary could be attached to the current release by [re-running all jobs](https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs) or the failed one.


kamiyaa:
Thanks!

PR #413, opened 2023-08-24
Bash script to install from the terminal

kennycallado:
There are a couple of things to improve:

The script currently has the version hardcoded. Perhaps using the [GitHub API](https://api.github.com/repos/kamiyaa/joshuto/releases) could be a better approach.

It's important to ensure compatibility with all attached versions in releases. Currently, it is functioning for the following platforms:
- x86_64-unknown-linux-gnu
- macOS versions from the terminal

I've included a function that allows for download verification. However, to use this function, it's necessary to generate the required file from __automated-build.yml__ and then simply uncomment two lines.

For the README, here are the installation instructions (I'm unsure where to place them):
```bash
# Basic installation
bash <(curl -s https://raw.githubusercontent.com/kamiyaa/joshuto/master/utils/install.sh)

# Specific release installation
# Replace 'v0.9.5' with the desired release version
RELEASE_VER='v0.9.5' bash <(curl -s https://raw.githubusercontent.com/kamiyaa/joshuto/master/utils/install.sh)
```
---


kennycallado:
Ups! I didn't realize about this, sorry. I'll open a new one.

PR #423, opened 2023-08-30
feat: Add rename_file_append_ext command

rqdmap:
This command renames a file with the cursor set to the end of the file name.

### DIFF
--- a/src/commands/rename_file.rs
+++ b/src/commands/rename_file.rs
@@ -69,25 +69,25 @@ pub fn rename_file_append(
     keymap_t: &AppKeyMapping,
 ) -> JoshutoResult {
     if let Some(file_name) = _get_current_file_name(context) {
-        let (prefix, suffix): (String, String) = match file_name.rfind('.') {
+        let (prefix, suffix) = (format!("rename {}", file_name), "".to_string());
+        command_line::read_and_execute(context, backend, keymap_t, &prefix, &suffix)?;
+    }
+    Ok(())
+}
+
+pub fn rename_file_append_base(
+    context: &mut AppContext,
+    backend: &mut AppBackend,
+    keymap_t: &AppKeyMapping,
+) -> JoshutoResult {
+    if let Some(file_name) = _get_current_file_name(context) {
+        let (prefix, suffix): (String, String) = match file_name.find('.') {
             Some(ext) => (
                 format!("rename {}", &file_name[0..ext]),
                 file_name[ext..].to_string(),
             ),
             None => (format!("rename {}", file_name), "".to_string()),
         };
-        command_line::read_and_execute(context, backend, keymap_t, &prefix, &suffix)?;
-    }
-    Ok(())
-}
-
-pub fn rename_file_append_ext(
-    context: &mut AppContext,
-    backend: &mut AppBackend,
-    keymap_t: &AppKeyMapping,
-) -> JoshutoResult {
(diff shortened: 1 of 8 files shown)


kamiyaa:
Good stuff!
But I think the name is a bit confusing.
It might be a better idea to swap the functionality of `rename_append` and `rename_append_ext`.

`rename_append` puts cursor to the end of file (appending to the file)
`rename_append_ext` puts cursor right before extension (appending to the extension)

Kind of scope creep, but what do you think?

rqdmap:
I mostly agree with you about these confusing names. It seems better to let `rename_append` act like the previous `rename_append_ext`.

But I want to reserve the ability to directly append to the file name without touching the extension. This functionality is quite frequently used I think. So I want to reserve it by naming it more reasonably. I did some research and found some naming methods for describing filename without extension:

- [Filename - Wikipedia](https://en.wikipedia.org/wiki/Filename) calls it `base name` or `stem`

	> Many file systems, including FAT, NTFS, and VMS systems, consider as filename extension the part of the file name that consists of one or more characters following the last period in the filename, dividing the filename into two parts: **a base name or stem** and an extension or suffix used by some applications to indicate the file type.

- Vim calls it `file root` (:help filename-modifiers).

I'm not sure which one is better(unambiguous and clear). I prefer `rename_append_base` because it seems simple and intuitive.

To sum up:

- `rename_append` puts cursor to the end of file;

- `rename_append_base` puts cursor to the end of the base name of file 

It looks good; What do you think of the new naming method?

kamiyaa:
`rename_append_base` is good for me :+1:

src/commands/rename_file.rs
```diff
@@ -69,7 +69,19 @@ pub fn rename_file_append(
...
+    Ok(())
+}
+
+pub fn rename_file_append_base(
+    context: &mut AppContext,
+    backend: &mut AppBackend,
+    keymap_t: &AppKeyMapping,
+) -> JoshutoResult {
+    if let Some(file_name) = _get_current_file_name(context) {
+        let (prefix, suffix): (String, String) = match file_name.find('.') {
```
rqdmap, line 84: Previously it is `rfind`, which will find the **last** matching pattern. I changed it to `find` to find the **first** pattern. It will be useful when handling file names with multiple suffix such as `a.tar.gz`:

```text
previous:	a.tar.gz
		    ⇑

now:		a.tar.gz
		⇑
```

kamiyaa: Not sure how I feel about this because sometimes files will contain periods in between that aren't necessarily related to the file extension.
It might be better to just `rfind` and `.tar.*` stuff just be categorized as a outlier.

For example, if a filename is a JWT token-like name 
```
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c.json
```
the cursor will be very far from the real extension.
At least with `rfind` the cursor is fairly close to the real extension (even for `.tar.*`)

rqdmap: You're right. I will change it back soon.

kamiyaa (approved):
LGTM!

Issue #425, opened 2023-09-01
utils directory


kennycallado:
I was updating the Dockerfile and moving it into a new **utils** directory, as I have done for the installation script. So, I was wondering if it is possible to move **nix** stuffs to keep the repo clean.

kamiyaa:
Maybe @Ruixi-rebirth can comment on this. I'm not too familiar with the nix system. But I imagine its not possible to move it?

Ruixi-rebirth:
> 

It is possible to move the nix directory to utils, I will make some small changes later. Until then you can do what you want

Ruixi-rebirth:
> I was updating the Dockerfile and moving it into a new **utils** directory, as I have done for the installation script. So, I was wondering if it is possible to move **nix** stuffs to keep the repo clean.

It is possible to move the nix directory to utils, I will make some small changes later. Until then you can do what you want

PR #427, opened 2023-09-06
Script to install from the terminal

kennycallado:
- Able to determinate the latest released version.
- Manage checksum file and ask in case there isn't.
- Should work for all platforms (macos, linux).

Update the automated-build workflow to generate the checksum files and remove deprecated use of set-output. Also add [profile.release] strip option in Cargo.toml to avoid extra step in automated-build.

Breaks _Nix Flakes_ workflow as mentioned in #425.

Shells tested:
- bash
- ash

Targets tested:
- aarch64-unkknown-linux-gnu
- x86_64-unknown-linux-gnu
- x86_64-unknown-linux-musl

Close: #425 and maybe #102


kamiyaa:
👍

PR #428, opened 2023-09-08
feat: extensions for select and filter

ctj12461:
Add a series of commands:

- `select_*`
  - `select_glob`
  - `select_regex`
  - `select_fzf`
- `filter_*`
  - `fzf_glob`
  - `fzf_regex`

BREAKING CHANGES: `select` now uses substring matching instead of globbing.


kamiyaa:
How does the `select_fzf` work?
Does it just select one item or multiple?

kamiyaa (approved):

PR #430, opened 2023-09-12
fix: add mutex lock to preview threads

luteran42:
Do not spawn 100s of preview threads when I'm holding the down arrow key in a crowded directory.


kamiyaa:
I'm okay with this being a temporary solution.
But will need to revisit this in the future and maybe come up with a better solution 👍

PR #431, opened 2023-09-12
feat: add option to use file_path() with the %p keyword

luteran42:
Use an absolute path where it is needed. like:
    { keys = ["m", "w"], commands = ["shell swaymsg output * bg %p fit"] },


kamiyaa:
LGTM!

Issue #434, opened 2023-09-13
preview_script required for preview_shown_hook_script?


EvanCarroll:
It's not at all clear to me that,

* if you don't have a `preview_script` that exists and is `+x` and returns 0, you can't get a `preview_shown_hook_script` working.
* why my `preview_removed_hook_script` doesn't work.

Now I know it, but I think the docs should reflect this very clearly.

DLFW:
Hi!
I will add a note in the docs. [Here](https://github.com/DLFW/joshuto/tree/doc-imgprev/docs/image_previews#image-thumbnails-in-file-previews) is a proposal (the ❗-section). Do you think that gives more clarity?
Sorry when it sound somewhat clumsy. I'm not an English native speaker (obviously). I'm always struggling with the documentation...

DLFW:
@EvanCarroll : can this be closed? Are you still missing something or do you have any proposal?

PR #435, opened 2023-09-19
Docs: Add note about text preview for img preview

DLFW:
The image-preview documentation now contains an extra, more explicit statement that the text preview pane must be shown for a file to trigger the shown-hook script.

Should solve #434.


kamiyaa:
Thanks!

PR #436, opened 2023-09-24
docs: add preview section to joshuto.yml example

radek-sprta:
Add preview section to joshuto.yml example configuration as it is defined in the default configuration. Using the example joshuto.yml as the base for custom configuration broke the preview functionality.


kamiyaa:
Thanks!

Issue #233, opened 2022-12-03
Hexadecimal representation of tabs 


rousbound:
![joshuto](https://user-images.githubusercontent.com/40251117/205461751-20c38cdc-48fe-4718-a801-c9b9a507a891.png)

When I try to build joshuto it always comes with this hexadecimal numbers instead of the tabs numbers on the right upper corner.

I've tried to build with "--release" and "--all-features" flags, but it always comes like this.

It's not a problem of the terminal, because with the pre-built joshuto that doesn't happen.

Does anyone have any ideas how to solve this?

Thanks in advance.

kamiyaa:
This is a new "feature" I guess.
Instead of tabs being stored in a vector, its being stored in a hashmap with ids.

I can however, add the numbers back if you would like

DLFW:
Ranger shows a tab-index and then the dir name. IMHO, that's the best option for the user to keep track what is opened where and how to activate each tab.
![Screenshot_ranger_tab_dir](https://user-images.githubusercontent.com/22507197/205516369-e9e7a842-6f33-495e-8f65-a0d4b247c738.png)

Would be great if we can have that for Joshuto. ;)

rousbound:
> This is a new "feature" I guess. Instead of tabs being stored in a vector, its being stored in a hashmap with ids.
> 
> I can however, add the numbers back if you would like

Well, I did not expect that xD. I personally found it a little bit distracting.

One option is the old one, which shows the tab as an Id out of the total tabs. 

Ex: "1/3"

The other one is range'rs with the example of the comment above.

Ex: "1: joshuto", "2: src"

The ranger's one is good because it identifies the directory the tab it is in, but in the other hand it becomes a like bit cluttered if your terminal window is short and you have lots of tabs.

Another option that we might consider is a 'carroussel-like' display:

Ex: |"9: /root", "1:joshuto", "2: src" |..... "3: /etc/passwd", "4:/var/log/apache2"....

Where only three are shown at a time, and the middle one is the current, so you would see PREVIOUS, CURRENT and NEXT tabs at a time. 

But these are just ideas, I don't use this feature very much actually, I mostly use bookmarks to jump around instead.

If it didn't show anything on the upper corner when you have just one tab it would be enough for me I think.

DLFW:
[...]

PR #437, opened 2023-09-30
Scrollable tab-bar

DLFW:
This implements a scrollable tab bar, like discussed in #233.

This makes tabs easier to read, better utilizes the space in the top-bar and makes handling many tabs way easier.

The display of the current directory has been removed. The current tab will show the directory always in full form. The directories of other tabs will be shown in full form if there is enough space, otherwise, they will be shown in a short form. If the available space in the top-bar is still not sufficient, scroll tags will be added at each end.

The long-form is currently defined as the full, absolute path. The short-form is currently defined as the last directory element of the path.

The tab-bar is configurable in terms of font styling and style character (like dividers, prefixes, and postfixes).

Documentation has been created in a separate file.

Collateral impact:
* The `tilde_in_titlebar` option for `[display]` in `joshuto.toml` has been removed. (Because the "current dir display" has been removed.)
* The `display_mode` option for `[tabs]` in `joshuto.toml` has been removed and with it, also the `tab_bar_mode` command. The appearance of index-numbers in tags is not configurable anymore. Numbers are shown if and only if there are more than one tab. This feature can be re-introduced later in some other form if still desired with the new tab-bar. With the new tab-bar, this configuration option did not make sense the way it was designed.
* The `max_len` option for `[tabs]` in `joshuto.toml` has been removed. (Because tabs don't have a fixed width anymore.)
* The function/feature of shortening paths by replacing heading path elements with their first character only has been removed. (Because I expect that this hard-to-interpret representation of a path will rarely be beneficial, now that there is a better space utilization in the top bar.)
* The "old" tab-bar configuration options have been removed.
* A `AppStyleOptionsRaw` struct has been introduced to allow handling of more complex default styling logic. Each styling attribute is defined there as an `Option` to indicate if a styling option has been set by the user or not.
* A `PathStyleIfSome` trait is now available in "utils" to patch ratatui styles with another, _optional_ style.


DLFW:
Quick link to the documentation for tab-bar configuration in my fork: https://github.com/DLFW/joshuto/tree/nicer-tabs/docs/configuration/tabbar

I would enhance the docs in some time, I guess.

Here are two screenshots of the default configuration; without scroll tags:
Screenshot_20230930_211141

...and with scroll tags:
Screenshot_20230930_211736

kamiyaa (approved):
Wow, this is amazing! Thank you!

kamiyaa:
Fixed crash :+1:  https://github.com/kamiyaa/joshuto/commit/cfcefaf9985adf766557e48db4fc2fb6d7498094

DLFW:
Thanks @kamiyaa! Sorry for the bug... I was testing this since weeks but it seems I'm not so interested in my root dir.... 😄

PR #439, opened 2023-10-03
feat: wezterm cwd compatibility

kennycallado:
This PR makes joshuto compatible with [wezterm](https://wezfurlong.org/wezterm/) at least with the way it manages the current directory updates.

1. Added a new `cwd` utility module.
2. Centralized CWD updates using the `cwd::set_current_dir` function.
3. Updated to use the `cwd` utility for CWD management.
4. Implements OSC 7: Escape sequences to inform the terminal about the current directory [ref](https://wezfurlong.org/wezterm/shell-integration.html#osc-7-escape-sequence-to-set-the-working-directory).

### DIFF
--- a/src/commands/change_directory.rs
+++ b/src/commands/change_directory.rs
@@ -4,10 +4,11 @@ use crate::commands::reload;
 use crate::context::AppContext;
 use crate::error::AppResult;
 use crate::history::DirectoryHistory;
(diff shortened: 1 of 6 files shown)


kamiyaa (approved):

kamiyaa:
Looks good

PR #444, opened 2023-10-09
fix: remove the debug info causing the chaos in UI

awerty-noob:
the redundent `eprintln` could cause UI unusable, so I comment it.


kamiyaa:
Thanks for the catch!

PR #446, opened 2023-10-14
custom_commands + fallback for the older configurations

Dzordzu:
1. Added custom_commands
2. Implemented custom_search
3. Implemented custom_serach_interactive
4. Added fallback for the command in the keymaps

Synchronized with the upstream version of the #339. Adds a fallback for the older keymap `command` parameter. It is translated into `commands` with single value. `command` takes precedence over `commands`

Also see #338 for reference


Dzordzu:
@kamiyaa I've done some sync. Let me know you've got any issues/suggestions to apply :)

Dzordzu:
@kamiyaa Just want to ping you. Before the codebase changes again ;)

superiums:
no custom commands need.
all shell command was custom commands.
the only thing need to concern was `pipe_out` and `pipe_in`
- pipe_out was done by `%s`
- pipe_in joshuto need to deel has only two:

1. dir to cd.
2. file/files to select.

superiums:
no custom commands need.
all shell command was custom commands.
the only thing need to concern was `pipe_out` and `pipe_in`
- pipe_out was done by `%s`
- pipe_in : there're only two thing joshuto need to deel :

1. dir to cd.
2. file/files to select.

maybe only need to read it from standerd output.

Dzordzu:
>  all shell command was custom commands

Not exactly. The purpose of the `custom_commands` is to provide some simple flexibility inside joshuto. It's a quick patch, that does not require massive code changes and refactoring. 

Honestly, I do not understand rest of your message.

superiums:
about fall back to old command, maybe there's a third way: use just old command, and add a separator such as `&&` or `;` to support multi command.

superiums:
> > all shell command was custom commands
> 
> Not exactly. The purpose of the `custom_commands` is to provide some simple flexibility inside joshuto built-in commands. It's a quick patch, that does not require massive code changes and refactoring.
> 
> Honestly, I do not understand rest of your message.

then, the define of custom command was  the first thing.
i guess the custom_command means custom search or sth like that.
if there is a way to make more flexiability of inner commands, that's very good.

i mean the `search_fzf`  `z`  `zi` or such an thing, may be able to be more flexable.
now there's `fd` replaced `find`, image years later, sth comes out to replace `z` `fzf`, our joshuto still support it without any coding update? 
or now, i wonder `rg` to be support in joshuto?
or now, i have an custom script to search out my files. how could i integrate it to joshuto ?

all these things can be done just with a litte restructure :
just receive piped back result after executed shell command. that's all.

for example: rg/fzf/z, they just do their own job. 
joshuto should not hard code them, but trait them as a common shell command.
and receive the result after execute.
- if the result piped back is a folder, just cd to the path in joshuto.
- if the result is a file, just select it. (if not current dir, cd to the path).
- if the result is many file/folder, just display them in one place. (just as search or flat command does)

this will be more flexable to all custom needs.

superiums:
in ranger, `shell -w` wait user to have a glance to the command result.
maybe we could:
`shell -p` to pipe the result back to joshuto.
`shell ` to run the command background and do nothing.
`shell -r` to run the command background and reload current dir in joshuto  .

kamiyaa (approved):

kamiyaa:
LGTM! Thanks!

PR #448, opened 2023-10-15
feat: add `focus_on_create` option

Akmadan23:
This is a nice feature that I felt was missing, I chose to enable it by default because in my opinion it just makes sense, but we can also change it to be `false` by default if you prefer. We can also change the option name, `focus_on_create` is the best I came up with, but if you think it can have a better and more clear name please let me know.


docs/configuration/joshuto.toml.md
```diff
@@ -20,6 +20,13 @@ use_trash = true
 # Watch for filesystem changes and update directory listings accordingly
 watch_files = true
 
+# If true the cursor will focus newly created files or directories with `:touch` or `:mkdir`
+# Even if true, the behavior can be avoided prefixing the new file/dir with "./"
+# E.g.:
+# - `:mkdir a` moves the cursor to the new directory `a`
+# - `:mkdir ./b` keeps the cursor where it was
+focus_on_create = true
```
Akmadan23, line 28: Hope this documentation is clear enough

kamiyaa:
LGTM! Thanks!

PR #451, opened 2023-10-20
Add thumbnail for video preview on kitty

Mroik:
Use `ffmpegthumbnailer` to generate a preview for videos just like ranger does


kamiyaa:
LGTM Thanks!

PR #483, opened 2024-01-23
fix overflow bug on bookmark widget when area is too small

pantosaur:
I found a bug where I get the following panic when there are too many bookmarks to fit inside the UI area:
```
thread 'main' panicked at src/commands/bookmark.rs:121:20:
attempt to subtract with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
The y coordinate of the ```ratatui::layout::Rect``` becomes -1 at line 121

I also made sure to truncate the TuiMenu options field to the maximum length, apparently ```area.height as usize - 1```, whenever it exceeds that value


kamiyaa:
LGTM Thanks!

Issue #486, opened 2024-01-24
Add a new command to print default configuration


sushi-shi:
### Description

Sometimes I want to get default config files without searching for them in the repository.
This can also be used by scripts to generate it.

PR #487, opened 2024-01-24
Add command for printing embedded configs

sushi-shi:
Resolves: #486 

A few things:
 * `get_type` trait function is not really needed, but I wanted to tie an actual config with its `ConfigType` value in some way.
 * I believe `include_str!("..\\..\\..\\..\\config\\joshuto.toml");` for windows is not needed, and `/` works just as well? I am not sure.
 * `get_config` can be moved out of a trait, but with this its usage will be uglier: `let app_config = get_config::<App>()`.
 * I thought about adding an option for printing the current config (might be useful for debugging), but let's discuss this first!

### DIFF
--- a/config/joshuto.toml
+++ b/config/joshuto.toml
@@ -23,7 +23,7 @@ show_icons = true
 # none, absolute, relative
 line_number_style = "none"
 
-# size, mtime, user, gourp, perm. can be combined with |. 
+# size, mtime, user, gourp, perm. can be combined with |.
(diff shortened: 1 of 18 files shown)


kamiyaa:
> I believe include_str!("..\\..\\..\\..\\config\\joshuto.toml"); for windows is not needed, and / works just as well? I am not sure.

Haven't tested but I guess we'll cross that road when we get there xd

Issue #443, opened 2023-10-09
[Feature Request] LS_COLORS support


linq:
Would it be possible to support LS_COLORS environment variable to set colors for files?

Thanks in advance

Beethoven-n:
make it a toggle so that existing themes still work

PR #489, opened 2024-02-11
feat lscolors support

benjaminedwardwebb:
# support `LS_COLORS`

This adds support for styling entries using the `LS_COLORS` environment variable. This styling is gated behind a configuration variable. If it's enabled, any other styling for entries based on theme configuration is ignored.

An example configuration line looks like

```toml
lscolors_enabled = true
```

in `theme.toml`.

This change adds a dependency on the [`lscolors`][0] crate. It addresses issue https://github.com/kamiyaa/joshuto/issues/443.

## verification

I verified these changes by running

```bash
nix develop --impure
, run
```

on my local machine.

```bash
uname -a
```
```
Linux berrio 6.1.72 #1-NixOS SMP PREEMPT_DYNAMIC Wed Jan 10 16:10:37 UTC 2024 x86_64 GNU/Linux
```

Here's an image of when `lscolors_enabled = true` in `theme.toml`.

![lscolors-enabled-true][1]

Here's an image of when `lscolors_enabled = false` in `theme.toml`.

![lscolors-enabled-false][2]

I also confirmed that

- when `lscolors_enabled` is not set in the configuration at all, `joshuto` behaves as if it is `false`
- older versions of `joshuto` ignore `lscolors_enabled` if it exists in the `theme.toml` file

[0]: https://github.com/sharkdp/lscolors
[1]: https://github.com/kamiyaa/joshuto/assets/7118777/e8241a6a-4462-4c72-b180-df61617ab904
[2]: https://github.com/kamiyaa/joshuto/assets/7118777/4f83550a-6599-443e-9c47-8c317b203ea1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -632,6 +632,7 @@ dependencies = [
  "globset",
  "lazy_static",
  "libc",
+ "lscolors",
  "nix 0.27.1",
  "notify",
  "open",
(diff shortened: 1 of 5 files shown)


benjaminedwardwebb:
It looks like this addresses an existing issue https://github.com/kamiyaa/joshuto/issues/443

kamiyaa (approved):
LGTM! thanks!

PR #492, opened 2024-02-14
fix: update all lists viewports on sort method change

Akmadan23:
I noticed that changing sort method in directories with many entries might not keep focus on the focused file (and the same goes for parent and child dirs). This patch fixes the issue.


kamiyaa:
Thanks!

Issue #402, opened 2023-08-14
[REQ] Make mouse support configurable


Sonico98:
Allow enabling / disabling mouse support from within joshuto.toml, with an option that allows setting mouse support to true or false. Right now, the only way to disable mouse support is to compile joshuto without the mouse feature.

kamiyaa:
Yeah, that would be a nice feature to have

Mroik:
This issue has been taken care of.

PR #494, opened 2024-02-15
Make mouse support configurable

Mroik:
Implements request of issue #402 for mouse support to be configurable through `joshuto.toml` instead of it being a cargo feature.


kamiyaa (commented):
Ideally I would like turning off mouse support to be able to select text.
As it is, when mouse is enabled or disabled, text cannot be selected via mouse.

So some programmatic way to use MouseTerminal

Mroik:
I really dislike this solution, made me change the signature of some functions to pass `context`. I don't think this is avoidable due to the fact that `Screen` was an alias had different types based on the `cargo features`.

kamiyaa:
> I really dislike this solution, made me change the signature of some functions to pass `context`. I don't think this is avoidable due to the fact that `Screen` was an alias had different types based on the `cargo features`.

An alternative solution could be to keep the mouse feature, have it enabled by default and can be changed via config (the work you did).
What do you think?

Mroik:
> An alternative solution could be to keep the mouse feature, have it enabled by default and can be changed via config (the work you did).

Not sure what you mean, as it is right now in this branch it is done only through the config (there wouldn't be a need for a cargo feature).
What I was referring to is [this](https://github.com/kamiyaa/joshuto/pull/494/commits/e48815d1d3b015f361ff1399e1c2be6f0785cb6f#diff-fe75a4d3e38caa85ca749984438fd437d8f87bfca931c0e6e3e330d744f72bf3R17). I don't think we can avoid an enum given that the previous implementation for the mouse support toggle via cargo feature achieved it through different types.

Mroik:
Not sure if I explained myself, sorry if I wasn't too clear :sweat_smile:

kamiyaa:
Okay, this is fine. I am content with this haha.

kamiyaa:
Thanks!

Issue #456, opened 2023-10-26
Add more parameter-variables to more commands (like `%s` for `shell`)


DLFW:
Hi,
to get on with https://github.com/kamiyaa/joshuto/issues/241#issuecomment-1546851335, I would maybe try to add more variables to more commands soon.

Currently, only `spawn` and `shell` support the variable `%s`, which is substituted by a list of names of the selected files.

The idea is...
1. ...to provide more variables (like `d` for the current directory, or `t` for a list of directories of all tabs )
2. ...to equip more commands with the ability to evaluate these variables
3. ...and to provide variables for the exit code and `stdout` of the last `shell` sub-process

1. would allow to hand over more information to external scripts, like the list of tabs, the highlighted file, or the current directory.

With 2. and 3. and the command-chains, introduced in 31beaab, it would be possible to process the output of scripts in joshuto, like opening a new tab in a directory that was calculated by a script, or to select certain files (would require a new command to select multiple files by index or name).

As we would have to format dynamic strings with a potentially growing number of named variables, this would require some kind of dynamic string formatting library or a template engine that is able to substitute sub-strings by name, not only by position. Because we don't want some complex HTML-template engine for this, I see only two options at hand in Rust-land:
* https://github.com/sunng87/handlebars-rust (A bit more complex but flexible and maintained)
* https://github.com/vitiral/strfmt (more lightweight but less maintained and using single `{}` for variables, which increases the chances for bad surprises for users)

If somebody knows a better alternative, please let me know.

However, all these solutions would require a breaking change in Joshuto: `%s` would change to  something else, for example `{{s}}`. @kamiyaa: Would that be acceptable?

I would probably implement the variable substitution in `src/key_command/impl_from_str.rs`, either for some dedicated commands or for the complete command-line string as it comes in. Have to think about it.

Input welcome. Maybe I would take this up in the next weeks.

DLFW:
Hi,
I'm not much further on this, I just was too busy with other stuff.

However, I did some experimental implementations around Christmas and I came to the conclusion that a generic variable solution I had in mind won't work well with the current way Joshuto parses the command line input. To get such a solution straight, it would require a complete refactoring of Joshuto's command handling. The problem is mostly with escaping and quoting (I don't even remember all the details). I thought of introducing some CLI parser library (like `clap`) to parse the commands, but that would be too much effort for me at the moment. In any case, it would have some negative impact on performance. (don't know how much; maybe negligible).

[...]

PR #495, opened 2024-02-15
Add `capture` and `stdout` commands

DLFW:
This adds two new commands as a base to enable users to use the output of scripts to do certain actions in Joshuto.

The first command this adds is a third command to start a sub-process beside `shell` and `spawn`, called `capture`. Like `shell`, `capture` is running blocking but unlike `shell`, it does not release the terminal but captures the `stdout` of the sub-process and stores it in an `AppContext` attribute.

The second command added by this commit is `stdout`. This command takes the output from the last `capture` run, stored in the `AppContext` attribute, and uses it for some action. The action has to be specified as a sub-command. As of now, only `stdout cd` is implemented. This checks that the last output of `capture` is a single line of an existing file or directory and then changes the working directory to that.

To get significant value from these new commands, `capture` needs to be equipped with more variables to feed more information about Joshuto's state into external scripts, and `stdout` needs to get some more sub-commands.


DLFW:
This is a first step for #456. Reasons why have chosen this solution are outlines [in my comment there](https://github.com/kamiyaa/joshuto/issues/456#issuecomment-1942655863).

As mentioned in the commit message, this is currently not sufficient to solve the original requirement, but I thought this is a good first commit, also to have some agreement that this concept is a feasible way to go. I will provide ~two follow-up commits to get some of the other missing pieces and I will also provide some documentation then.

To test the single working “stdout-command”, one can add something like 
```yaml
  { keys = ["e", "t"], commands = ["capture joshtest.sh", "stdout cd"] },
```
into `keymap.toml`, a script `joshtest.sh` like
```bash
#!/usr/bin/env bash
echo $(cat /home/dude/test-bookmarks | dmenu)
```
somewhere in the path, and a list of directories in `/home/dude/test-bookmarks` like
```
/home/dude/Downloads
~/music
~/sandbox/joshuto
```

Then, `e-t` will open a dmenu where one can choose a directory from the list, and Joshuto will jump to there.

It's just a first, simple example for these two new commands.

DLFW:
Just did a rebase and fmt

kamiyaa:
Thanks! Might take a while for me go through the PR

DLFW:
Just rebased and resolved conflicts

kamiyaa:
Interesting design.
It seems a little messy to have these commands do all this, but I can't think of a better solution at the moment and I'm curious to see where this goes.

kamiyaa (approved):

DLFW:
> Interesting design. It seems a little messy to have these commands do all this, but I can't think of a better solution at the moment and I'm curious to see where this goes.

“A little messy” is a good description. 😇  In the end, I saw only two alternatives: Complete refactoring of the “command processing” in Joshuto and introducing something like `clap` to have a real command/option/argument parser (which would have enabled me to implement my initial idea), _or_ introducing something like a socket from where Joshuto can read commands (in parallel to stdout) to allow “remote controlling” of Joshuto. That could have also been used by `shell`-commands to issue certain reaction. Both these alternatives were too much effort/risky for me right now.

Thanks for merging!

PR #496, opened 2024-02-16
fix selection style when lscolors enabled

luteran42:
Selection style is not applied when the "lscolors" option is enabled.


kamiyaa (approved):

kamiyaa:
Thanks!

PR #497, opened 2024-02-22
feat: add `case_sensitive_ext` option for mimetype, theme and icons

Akmadan23:
Currently any file with an uppercase or partially uppercase extension is not recognized by either `mimetype.toml`, `theme.toml` and `icons.toml`. With this addition, when the flag is set to `false`, any uppercase/lowercase variant of an extension will be correctly detected. I set it to `false` by default because it just makes sense to me, but we can set it to `true` if we want to keep the current behavior by default.


src/main.rs
```diff
@@ -68,6 +68,8 @@ lazy_static! {
 
         config_dirs
     };
+
+    static ref CONFIG_T: AppConfig = AppConfig::get_config();
```
Akmadan23, line 72: Having it as a static reference here seems to be the only solution... That's not a problem right?

kamiyaa: Instead of making the config static, I think the less intrusive approach is to just check the config to see if option is enabled and pass the already lowercase version into `app_list_for_ext`.

ie. inside `open_file.rs`, change `_get_options` to take a `&AppConfig`

Akmadan23: Well, that works but just for the mimetype part, for theme and icons there is something more to do, but I'll try

Akmadan23:
Here is the new implementation, which involves a lot of extra clones, not sure if that's better than having a static config.

Speaking of clones, I was wondering if it would be a good idea to do a major refactor using only one static config (like `THEME_T`, `MIMETYPE_T`, etc.) and keeping in the context just a minimal config to be used only for mutable values. It would drastically decrease clones and improve both memory and cpu performance. Let me know your thoughts on this...

kamiyaa:
> Here is the new implementation, which involves a lot of extra clones, not sure if that's better than having a static config.
> 
> Speaking of clones, I was wondering if it would be a good idea to do a major refactor using only one static config (like `THEME_T`, `MIMETYPE_T`, etc.) and keeping in the context just a minimal config to be used only for mutable values. It would drastically decrease clones and improve both memory and cpu performance. Let me know your thoughts on this...

Yeah, the issue is with static, its read-only I believe.
So we won't be able to modify the configs (ie. show_hidden).
I guess we can get around this with an `Arc<>` or something

kamiyaa:
I will experiment with moving config into a static mutable ref and see how it is.

PR #498, opened 2024-02-22
feat: use a pager instead confirm_exit

luteran42:
Read the $PAGER environment variable. If it's not set, simply use 'less'.


kamiyaa (approved):
Looks good to me :+1: 
Didn't realize thats what people used for this xd

PR #500, opened 2024-02-22
feat: add marking and cancelling for file operations

luteran42:
Implement marking functionality and the ability to clear marks,
allowing for the cancellation of file operations.

also deselects entries after a file operation succeeds.

key:
`{ keys = ["alt+v"], commands = ["cancel_file_operation"] },`

e.g:
2024-03-14T01:11:36,900766630+01:00

alt+v:
2024-03-14T01:11:45,752700310+01:00

I like getting visual feedback, when I yy/pl/x files. 🙂

Update: also different colors for file operations:
```toml
[mark.cut]
fg = "red"
bold = true
# prefix = " "

[mark.copy]
fg = "rgb(235, 104, 65)" # orange
bold = true
# prefix = " "

[mark.symlink]
fg = "rgb(254, 132, 172)" # pink
bold = true
# prefix = " "
```


config/theme.toml
```diff
@@ -38,12 +38,16 @@ bold = true
...
+fg = "light_green"
 bold = true
 
 # Files selected in current visual mode
 [visual_mode_selection]
-fg = "light_red"
+fg = "light_yellow"
+bold = true
+
+[mark]
```
Akmadan23, line 49: Wouldn't it be nice to have different highlight groups for different file operations? Maybe something like `[mark.copy]`, `[mark.cut]` and `[mark.link]`

kamiyaa:
While I like the idea of marking files that have been cut/copy into "clipboard",
I'm finding the solution a bit cumbersome ergonomically.

I will need to do some more testing to figure out what it is, will keep you posted :+1:

scutuatua-crypto (approved):

scutuatua-crypto (approved):

scutuatua-crypto (approved):

PR #501, opened 2024-02-22
fix: escape `'` char in trash operations

Akmadan23:
Without this any attempt to trash a file with a `'` in its path results in an error.


kamiyaa (approved):
lgtm! I kind of knew of this issue but didn't bother fixing it xdd

Issue #180, opened 2022-06-26
Better support for autocompletion


sushi-shi:
Currently auto-completion is very simple, it expects the first word to be a command and next words to be files. This is already wrong, as some commands are made of multiple words (`sort reverse`) and what's more some commands do not accept files as parameters to begin with, so we do not need auto-completion for them. And other commands accept only folders, so autocompleting files should be avoided.

I propose to either add a simple specification, which every command should implement, to aid with autocompletion.
Something akin to:
```rust
struct Command {
    spec: Specifiaction
    ...
}

struct Specification {
    params: Vec<(&'static str, Specification)>, // to cover `sort reverse`, `cd ..` and `cd -`.
    file_ty: Option<(FileTy, usize)>, // usize to show how many files the command might accept
}

enum FileTy {
    Any,
    Directories,
    FIles, 
}
```
This specification is in no way complete, but it covers the most general cases, which I think should suffice for now.
Or make `Command` implement autocompleting interface through traits. (Possibly with a default implementation).

After we decide what to do (and whether to do anything at all), I will be willing to implement it.

sushi-shi:
Also `open_with` requires current programs in PATH for auto-completion

kamiyaa:
I agree that autocompletion is much more work. :+1: 

I'm kind of busy at the moment, but I'm definitely open to discussing how we want to go about designing and implementing this. :)

PR #502, opened 2024-02-28
fix: path autocompletion

Akmadan23:
This doesn't implement #180 (which I'd like to start working on), but at least makes path autocompletion usable.


kamiyaa (approved):
LGTM, thanks!

PR #506, opened 2024-03-11
Zoxide related features

### LINKED ISSUE #240: zoxide can't access directories it hasn't indexed
Trying to ```z``` inside a directory that zoxide has never accesed before returns ```zoxide: no match found```. This prevents replacing cd with zoxide, as commands like ```z ..``` won't work.

To test, outside of joshuto, create 2 directories and access one of them with zoxide
```
mkdir /tmp/accessed_directory /tmp/not_accessed_directory
z /tmp/accessed_directory
```
Then, open joshuto and try browsing to each of them. Only the accessed directory works
```
:z /tmp/accessed_directory --> Works
:z /tmp/not_accessed_directory --> zoxide: no match found
```
In the same fashion, trying to access sub-folders on the home directory with ```~``` doesn't work, while accessing them with their absolute path works
```
:z /home/user/Downloads --> Works
:z ~/Downloads --> zoxide: no match found
```

joshuto 0.9.4 installed from source as of commit 5ea87f65df0b0744ee1d01fd40e240e6abf0da77

### LINKED ISSUE #395: [FR] Support multiple patterns with `zoxide` (interactive)
Basically https://github.com/kamiyaa/joshuto/issues/219 but with `zi` (interactive).

In Zsh, `zi <pattern>` only prints paths that match `pattern` 

Currently, `:zi <pattern>` in joshuto is the same than `:zi`, all paths from `zoxide` database are printed.

### LINKED ISSUE #452: how to use cmd_aliases for command with user input?
in `joshuto.toml` I configed 

```toml
[cmd_aliases]

j = "z"
```

Expected `j` could be used as same as `z`, but `cmd_aliases` was not able to take the shell input from user and used for zoxide.

Is there a way to replace a command?🙏

### LINKED ISSUE #459: Track File Path History in Joshuto and Update Zoxide Database
How to track file path visit history in Joshuto and update Zoxide database? 
For instance, when creating a new directory and then navigating to it in Joshuto, it appears that this action isn't reflected in Zoxide, as 'z <new_dir_name>' results in an error: 'zoxide: no match found'.

chrjabs:
This pull request implements some features related to zoxide:
- Make `:z` behave more like zoxides `z` command: in the first place, act like `:cd`, only if this behaviour fails, query the zoxide database and naviagate to the found match (resolves #240)
- Introduced an opt in feature that updates the zoxide database after every directory change (notably also with manual navigation or `:cd`). The configuration option is called `zoxide_update`. (resolves #459)
- Support for arguments passed to the zoxide interactive (`:zi`) command (resolves #395)
- Support for command aliases with arguments (note: only aliases that do not contain spaces support arguments) (resolves #452)

One question related to the debug assertion introduced [here](https://github.com/chrjabs/joshuto/blob/aaea0c612fb4d623d154336c36344b3962c65764/src/commands/change_directory.rs#L14): this is required for the zoxide update to work correctly, but will this always hold? I am confident that this holds when coming from [`change_directory::change_directory`](https://github.com/chrjabs/joshuto/blob/aaea0c612fb4d623d154336c36344b3962c65764/src/commands/change_directory.rs#L34), but the function is also called from [`open_file::open`](https://github.com/chrjabs/joshuto/blob/aaea0c612fb4d623d154336c36344b3962c65764/src/commands/open_file.rs#L165), where I am not entirely sure.


kamiyaa:
> function is also called from [open_file::open](https://github.com/chrjabs/joshuto/blob/aaea0c612fb4d623d154336c36344b3962c65764/src/commands/open_file.rs#L165), where I am not entirely sure.
This one should be fine because its taken from `entry.file_path()`

kamiyaa (approved):

PR #507, opened 2024-03-13
feat: add `--reverse` flag to sort methods

Akmadan23:
This can be useful to get to a specific order with just a command without the need of toggling reverse each time, for example when I switch to `mtime` sort I often want newer files on top, which is reversed.


kamiyaa:
LGTM Thanks!

PR #511, opened 2024-03-14
fix: restore line numbers

Akmadan23:
I noticed that the eager evaluation of prefix width introduced in 1822104 ignores the line number part, which is pushed to prefix later.


kamiyaa:
Oops!

kamiyaa (approved):

PR #513, opened 2024-03-15
chore(nix): update `flake.lock`

xrelkd:
Update `Nix flake` to solve the problem: https://github.com/kamiyaa/joshuto/actions/runs/8303354942/job/22727337031#step:4:1004

This allows us to use a newer `rustc` while compiling `joshuto`.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -8,11 +8,11 @@
         "pre-commit-hooks": "pre-commit-hooks"
       },
       "locked": {
-        "lastModified": 1693363254,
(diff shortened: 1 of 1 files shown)


kamiyaa:
Thanks! Yeah I wasn't sure why it was failing hmm

xrelkd:
You may use
```
# update nix flake
nix flake update --commit-lock-file --commit-lockfile-summary 'chore(nix): update `flake.lock`'  

# build joshuto with nix
nix build

# test the executable
./result/bin/joshuto version
```
to update and test the Nix flake.

PR #515, opened 2024-03-19
feat: add support for %s and %p as first argument to subprocess commands

Akmadan23:
This allows to run the file under cursor (if executable) with `:shell %s`, `:shell ./%s` or `:shell %p`.

I also fixed a bug introduced in #495 which caused a panic when trying to run a file that doesn't exist or a non-executable file.


kamiyaa (approved):

kamiyaa:
Looks good, thanks!

PR #520, opened 2024-04-02
file operations export to toml file

pantosaur:
I made an experiment to be able to cut/copy/symlink files across multiple instances of joshuto. I find this useful when, let's say, i am working on multiple projects at the same time, and using multiple file browsers. My file browser is generally not my workflow starting point, but rather a tool I use inside my workflow, so I'm rarely gonna have a joshuto tab for everything I am doing, hence, my need for this. 

I used serde to serialize the file operation to a toml file in .config/joshuto/state_context.toml, and deserialize the file content when pasting the result. So far this seems to be working. I tried to follow the code structure as much as possible, but it could be improved. And there might be some considerations about the general approach. But in any case, this is a feature I want, so I am leaving this here as a general idea for consideration.

Thank you!


DLFW:
Hi! This sounds really useful! I have the same problem: I use many “Joshutos” all over and sometimes it would be very helpful to move or copy files over several instances. As of now, I always copy the path and then either open another tab in another Joshuto or quickly jump into the shell. However, just have copy/paste/link working cross-instances would be amazing.

I was thinking, all these standard file manager (pcmanfm, dolphin, thunar,  u name it...) can use the clipboard to copy or move (and maybe even symlink?) files. How do they do that? Maybe we could use `xclip` / `wl-clipboard` to do this? Would have the benefit of even being interop with other software. 

I didn't find a proper spec how such file operations are done with the clipboard. Does somebody know more about this?

pantosaur:
@DLFW I used to do this, yank path, and call a shell command from joshuto that pastes from the clipboard, but it is not very clean to use the clipboard for this, and more importantly, it is not a proper way to move around multiple files with a decent format to represent and store all the information about the transaction, i.e. copy, cut, symlink, list of files, etc. So, since the crate is already using serde+toml, it is fairly easy to do it in rust directly rather than external scripts, although we could also compile an external rust program to do just this.
The more involved and probably better solution rather than using a file to store the command, would be to make a custom process to handle this for all open file browsers, or use dbus, which I think is made pretty much for that purpose, and might already have a standard for exactly this.

scutuatua-crypto (approved):

scutuatua-crypto (approved):

scutuatua-crypto (approved):

scutuatua-crypto (approved):

scutuatua-crypto (approved):

PR #525, opened 2024-04-08
fix: hide image preview in fzf and task view

Akmadan23:
I noticed that switching to fzf or task view with an image preview visible, it remains there covering part of the view, so this fixes the issue.


kamiyaa (approved):
Looks good :+1:

PR #526, opened 2024-04-09
feat: comprehensive autocompletion for each command

### LINKED ISSUE #180: Better support for autocompletion

Akmadan23:
As anticipated in #502, this is my implementation of a comprehensive autocompletion for each command (closes #180), which can be of 4 kinds:
- Any file
- Directories (`cd`, `new_tab`)
- Executable files (`shell`, `spawn`)
- A custom list of arguments or flags

More details in the comments below.

### DIFF
--- a/src/key_command/constants.rs
+++ b/src/key_command/constants.rs
@@ -1,4 +1,4 @@
-use rustyline::completion::Pair;
+use lazy_static::lazy_static;
 
 pub const CMD_COMMAND_LINE: &str = ":";
 
@@ -8,8 +8,8 @@ macro_rules! cmd_constants {
             pub const $cmd_name: &str = $cmd_value;
         )*
 
-        pub fn commands() -> Vec<&'static str> {
-            vec![$($cmd_value,)*]
+        lazy_static! {
+            pub static ref COMMANDS: Vec<&'static str> = vec![$($cmd_name,)*];
         }
     };
 }
@@ -98,14 +98,3 @@ cmd_constants![
     (CMD_CUSTOM_SEARCH, "custom_search"),
     (CMD_CUSTOM_SEARCH_INTERACTIVE, "custom_search_interactive"),
 ];
-
-pub fn complete_command(partial_command: &str) -> Vec<Pair> {
-    commands()
-        .into_iter()
-        .filter(|command| command.starts_with(partial_command))
-        .map(|command| Pair {
-            display: command.to_string(),
-            replacement: command.to_string(),
-        })
-        .collect()
-}
--- a/src/key_command/impl_completion.rs
+++ b/src/key_command/impl_completion.rs
@@ -1,0 +1,72 @@
+use super::constants::*;
+use super::{Command, Completion};
+
+pub enum CompletionKind<'a> {
+    Bin,
+    Custom(Vec<&'a str>),
+    Dir(Option<Vec<&'a str>>),
+    File,
+}
+
+impl Completion for Command {
+    fn completion_kind(cmd: &str) -> Option<CompletionKind> {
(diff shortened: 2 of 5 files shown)


src/key_command/constants.rs
```diff
@@ -8,8 +8,8 @@ macro_rules! cmd_constants {
             pub const $cmd_name: &str = $cmd_value;
         )*
 
-        pub fn commands() -> Vec<&'static str> {
-            vec![$($cmd_value,)*]
+        lazy_static! {
+            pub static ref COMMANDS: Vec<&'static str> = vec![$($cmd_name,)*];
```
Akmadan23, line 12: This isn't a necessary change, but this way the array gets initialized just once instead of on each call.

src/key_command/impl_completion.rs
```diff
@@ -0,0 +1,72 @@
...
+
+pub enum CompletionKind<'a> {
+    Bin,
+    Custom(Vec<&'a str>),
+    Dir(Option<Vec<&'a str>>),
+    File,
+}
+
+impl Completion for Command {
+    fn completion_kind(cmd: &str) -> Option<CompletionKind> {
```
Akmadan23, line 12: I initially tried to match on `self`, but some commands need specific conditions to be successfully built via `from_str`, so matching just on the keyword seems to be the only solution. So this could have been a standalone function in `ui/view/tui_textfield.rs`, but I think that having it here as a trait of `Command` makes it more discoverable for future contributors.

src/ui/views/tui_textfield.rs
```diff
@@ -13,11 +17,49 @@ use unicode_width::UnicodeWidthStr;
...
 use crate::event::AppEvent;
-use crate::key_command::{complete_command, Command, InteractiveExecute};
+use crate::key_command::constants::COMMANDS;
+use crate::key_command::{Command, Completion, CompletionKind, InteractiveExecute};
 use crate::ui::views::TuiView;
 use crate::ui::widgets::{TuiMenu, TuiMultilineText};
 use crate::ui::AppBackend;
 
+lazy_static! {
+    static ref EXECUTABLES: Vec<&'static str> = {
```
Akmadan23, line 27: This generates the list of binaries in `$PATH`. Should work well both on unix and windows (I tested it on linux, can't confirm about windows).

src/key_command/impl_completion.rs
kamiyaa: Yeah, this looks great :+1:

src/key_command/constants.rs
kamiyaa: Cool, I didn't know you could do that haha

kamiyaa (approved):

PR #537, opened 2024-05-06
XDG-thumb for preview & thumbs for non-image files

DLFW:
This implements a cache for thumbnails and image previews for non-image file formats, like videos and PDFs.
This patch extends the inbuilt image preview feature (285df85) by integrating the `allmytoes` crate and using its thumbnails instead of the actual file being previewed.

The image-preview feature uses the XDG-(freedesktop.org-) specified thumbnail-cache to re-use existing thumbs and to store newly created thumbs.
(https://specifications.freedesktop.org/thumbnail-spec/thumbnail-spec-latest.html)

This increases performance for the image-preview because the preview does not need to load the full-sized version of the image and does not need to scale it down every time, once the thumbnail has been created. Also, thumbnails are now shared with other programs that use the XDG-thumb cache.

Furthermore, the new way to obtain thumbs allows to show image-previews for file types other than images. Existing thumbnails are shown for any file type. New thumbnails can be created for many video formats, PDF files, Postscript files, and SVGs. Users can also add “providers” for other file-types by configuring `allmytoes`.

The XDG-thumb feature can be disabled, and the thumbnail-size can be changed, both in `joshuto.toml`. The documentation has been extended, the whole image-preview page has been enhanced a little.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,6 +30,34 @@ dependencies = [
 ]
 
 [[package]]
+name = "allmytoes"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1c7b3df50220b4e21b9923db076437d6fc7509d2aa7219b975e8545c1bacaf1c"
+dependencies = [
+ "byteorder",
+ "clap",
+ "dynfmt",
+ "flexi_logger",
+ "hex",
+ "image",
+ "imagesize",
+ "kamadak-exif",
+ "log",
+ "md5",
+ "mime",
+ "once_cell",
+ "pathdiff",
+ "percent-encoding",
+ "png",
+ "serde",
+ "serde_yaml",
+ "subprocess",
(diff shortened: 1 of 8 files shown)


DLFW:
**Disclaimer**: I'm the author of [AllMyToes](https://gitlab.com/allmytoes/allmytoes).

Two notes on this:

* AllMyToes is currently GPL licensed. As I'm the author, I sill dare to provide this PR, even though Joshuto is LGPL. 🙂 I will eventually change the license soon, most probably to MIT.
* I have never tested AllMyToes on Mac and I have no idea if that works. If MacOS does not use a "magic" database for mime-types like Linux does, it will probably not work. Mac users would need to switch off this feature. If there are Mac users around, I would appreciate some feedback.

kamiyaa (approved):
Wow, this is really neat!
AllMyToes is a really cool project and I'll probably use it myself lol.

PR #552, opened 2024-07-20
configure preview image background color via theme.toml

### LINKED ISSUE #521: Configure ratatui-image preview background/frame color.
https://github.com/benjajaja/ratatui-image/issues/21

image
Due to ratatui's immediate-mode nature, the cells behind the image must be "skipped" so that nothing draws over the image. In other words, the cells behind the image still have whatever was drawn last time before they were marked skipped for the image. Thus, the image must always fill the cells/area, and the frame cannot be made transparent.

**The background color should be configurable.**

benjajaja:
https://github.com/benjajaja/ratatui-image/issues/21

image
Due to ratatui's immediate-mode nature, the cells behind the image must be "skipped" so that nothing draws over the image. In other words, the cells behind the image still have whatever was drawn last time before they were marked skipped for the image. Thus, the image must always fill the cells/area, and the frame cannot be made transparent.

This PR makes the background color configurable via theme.toml. It could also be done via `joshuto.toml`, which is what I initially did, so I could revert to that. I am not sure which is better, the goal here is essentially to make previews look good in terminals with white background, but it actually also applies to any terminal theme background color that isn't pitch black.

Closes #521

### DIFF
--- a/config/theme.toml
+++ b/config/theme.toml
@@ -2,6 +2,8 @@
 ## General
 ##########################################
 lscolors_enabled = false
+# The image preview background color, only RGB colors are accepted.
+# preview_background = "#FFFFFF"
 
 ##########################################
 ## Tabs
--- a/docs/image_previews/README.md
+++ b/docs/image_previews/README.md
@@ -21,6 +21,8 @@ preview_protocol = "halfblocks"
 ...
 ```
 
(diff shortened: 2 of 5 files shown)


kamiyaa (approved):
Looks good

PR #553, opened 2024-07-20
update ratatui-image to 1.0.5

benjajaja:
Should bring in some improvements like tmux support.

Also align `image` dependency to `allmytoes`'.


kamiyaa (approved):
LGTM!

PR #565, opened 2024-09-16
remove ctime for musl environment

usamoi:
#549 introduces "ctime" linemode but it makes joshuto immediately panic because it calls `Metadata::created`. For `x86_64-unknown-linux-musl` target, it always panics.

This PR removes "ctime" temporaily.

You can merge this PR and revert this PR when `Metadata::created` is supported for musl in Rust.


kamiyaa:
I think what we can have is a default SystemTime value (like default unix epoch time or something) when compiling to musl target

How does that sound?

usamoi:
> How does that sound?

If they notice that all their files were created on 1970-01-01, wouldn't they feel confused?

kamiyaa (approved):

PR #571, opened 2024-10-26
Add minimal display mode

flipsi:
This adds a new display mode referred to as `minimal`. 

[It does exactly what it says on the tin](https://en.wikipedia.org/wiki/Does_exactly_what_it_says_on_the_tin) - basically the current tab content is rendered full size (no miller columns like in `default`, no multipanes like in `hsplit`).

Screenshot_2024-10-26_22-59-44

This is similar to the `multipane` view in ranger when there is only one tab.

Disclaimer: I was personally missing this feature and built it by copy-pasting and cutting down existing code - not really knowing Rust. I hope I didn't miss anything important.

### DIFF
--- a/src/ui/views/tui_view.rs
+++ b/src/ui/views/tui_view.rs
@@ -4,7 +4,7 @@ use ratatui::widgets::Widget;
 
 use crate::types::option::display::DisplayMode;
 use crate::types::state::AppState;
-use crate::ui::views::{TuiFolderView, TuiHSplitView};
+use crate::ui::views::{tui_minimal_view::TuiMinimalView, TuiFolderView, TuiHSplitView};
 
 pub struct TuiView<'a> {
     pub app_state: &'a AppState,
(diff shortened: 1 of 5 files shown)


src/ui/views/tui_view.rs
```diff
@@ -4,7 +4,7 @@ use ratatui::widgets::Widget;
 
 use crate::types::option::display::DisplayMode;
 use crate::types::state::AppState;
-use crate::ui::views::{TuiFolderView, TuiHSplitView};
+use crate::ui::views::{tui_minimal_view::TuiMinimalView, TuiFolderView, TuiHSplitView};
```
flipsi, line 7: I don't understand why I had to qualify this to make it compile, any help/explanation is appreciated.

kamiyaa: if you add

```rust
pub use self::tui_minimal_view::*;
```

then it won't require you to do this. Up to you on if you add it though, I don't have a huge preference and will probably change it later anyways xd

kamiyaa (approved):
Looks good :+1:

src/ui/views/tui_view.rs
flipsi: Thanks for the hint!

PR #572, opened 2024-10-26
Set display mode command

flipsi:
This adds a command `set_display_mode` to change the display mode at runtime.

Note my other PR #571 which this PR is based on, so the other should be reviewed and merged first for a cleaner diff. Thanks!


kamiyaa (approved):
Looks good :+1:

PR #577, opened 2024-11-05
refactor: replace manual sign detection with format spec "+"

Integral-Tech:
- According to [the documentation of std::fmt](https://doc.rust-lang.org/std/fmt/index.html#sign0):
> *\+* - This is intended for numeric types and indicates that the sign should always be printed. By default only the negative sign of signed values is printed, and the sign of positive or unsigned values is omitted. This flag indicates that the correct sign (+ or -) should always be printed.


Integral-Tech:
@kamiyaa CI fails on the main branch due to code format. I have fixed it in my PR :)

PR #591, opened 2024-12-10
update default foreground colour for files to be compatible with different terminal themes

### LINKED ISSUE #340: Default color for regular files
Now it set to `white`: https://github.com/kamiyaa/joshuto/blob/main/config/theme.toml#L17-L18
Which makes the files invisible on white background and probably hardly visible in most light terminal themes. Would you consider not changing this color?

j-lakeman:
possibly closes #340


kamiyaa (approved):
lgtm, thank!

PR #609, opened 2025-06-27
Update `quit` integration examples

huskyistaken:
The example provided is bash specific, I thought it might be better to have it be more generic.

I also added a fish specific snippet.


kamiyaa (approved):
Looks good, thanks!

PR #618, opened 2025-12-26
feat: add navigation history

Akmadan23:
This is a per-tab implementation of navigation history in the style of what ranger and others already have. For the end user there are two new commands: `history_next` and `history_prev` which do exactly what the name says, pretty straightforward.


scutuatua-crypto (approved):

kamiyaa (approved):
Thank you!

